PR #27, opened 2013-12-11
Add Drupal to CMS category

jief:
Add Drupal to CMS category


darol100:
Drupal 8 was out in Nov 19, 2015. I think we should added Drupal into a list.

I also think we should  add a "CMS" section that can contain all php cms.

ziadoz:
@darol100 If you make a pull request for a new CMS category I'll take a look.

PR #150, opened 2014-07-08
Added FatFree micro framework

elvendor:


MissAllSunday:
I was about to submit a PR for F3  :P 

Anyway, I think F3 would be a nice addition, they do some things quite different than other frameworks  so for the sake of plurality it would be nice to add them. Its a full framework rather than a micro framework though,

It does have tests,  they have their own test environment built in. As weird as that sounds :tongue: 

So, can this be revised again or is it set on stone to decline it?

I would also recommend lithium although it seems it was added and later removed.

Greetings.

sn0opy:
Since this is the only Issue / PR with an official statement why it hasn't been added yet, I'd like to draw some attention back to it to get it finally into the list.

In the latest run, it's (one of) the fastest PHP frameworks: https://www.techempower.com/benchmarks/

Add it to the framework section, if you think it's not micro enough.

PR #239, opened 2014-09-30
Added script for checking formatting and URLs.

h4cc:
The script will check the format of each listItem. There are currently a few exceptions, where multiple links to blogposts series are given.

The script is also checking all Urls it can find in the document and check these for HTTP StatusCode ~200. Only owasp.org had to be excepted, because curl does not like them (?).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: php
+
+php:
+   - 5.6
+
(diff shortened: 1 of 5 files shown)


josegonzalez:
Already done in master using `awesome_bot`. Thanks though!

PR #253, opened 2014-10-22
Adding Broadway to the list of event driven libs

sdepablos:
Broadway, an event source and CQRS library for PHP https://github.com/qandidate-labs/broadway


ziadoz:
Thanks! I've merged this.

PR #255, opened 2014-10-26
Add PhpDependencyAnalysis for Code Analysis

mamuz:
PhpDependencyAnalysis is an extandable static code analysis to provide a dependency graph for abstract datatypes based on namespaces. It creates graphs on customizable levels, e.g. on package-level or on class-level.


mamuz:
Since v0.3.0 this tool supports also verifing created dependency graph against a defined architecture as a key feature.
All violations like cycles and references which are not matched to defined architecture will be highlited.

ziadoz:
Thanks, this looks great.

PR #260, opened 2014-10-30
docopt.php

jkufner:
docopt creates beautiful command-line interfaces. Command line argument parser, uses documentation to specify options.

https://github.com/docopt/docopt.php


ziadoz:
Thanks for the suggestion.

PR #266, opened 2014-11-16
Zend Cache

marc-mabe:
The Caching Library of Zend Framework 2

PS: Also prefixed the caching library of Doctrine.


andytruong:
I got 404 for https://github.com/zendframework/Component_ZendCache
I think the correct link is https://github.com/zendframework/zend-cache

PR #271, opened 2014-11-25
Middlewares

ojhaujjwal:
I added a new `Middlewares` section. Middlewares in PHP are getting popular these days because of evolution of NodeJs's Express and Connect.


ziadoz:
This is a great change. I've been wanting to breakout middlewares into their own category for a while now. Thanks.

PR #273, opened 2014-11-25
oauth1 & oauth2 client libraries from thephpleague

ojhaujjwal:


ziadoz:
Thanks, I've merged this.

PR #277, opened 2014-12-04
Mocking library which solves the method chaining mocking diffculties

danrevah:
I'd like to nominate a PHP Mocking Framework inspired by Java Mockito library (support from PHP 5.4+). which solves the Method Chaining difficulties when mocking chained methods.
and also contains partial mocks 
https://github.com/danrevah/ShortifyPunit
install able by Composer and PSR-4 compilant
100% Code coverage, high code quality/standards and is actively developed


ziadoz:
Thanks for the suggestion.

PR #279, opened 2014-12-05
Added thephpleague's oauth2-server to the Authentication library list

edigu:


ziadoz:
Thanks. This is already the list, it just had the League's old URL. I've corrected that, and the links for their other packages too.

PR #280, opened 2014-12-10
String Sequence Incrementor 'SSI'

gitall:
SSI is a PHP based library which provides a sequential increment in an alpha numeric string order


ziadoz:
Thanks for the suggestion.

PR #281, opened 2014-12-10
String Sequence Incrementor

gitall:
Repository name has been changed & updated


ziadoz:
Thanks for the suggestion.

PR #282, opened 2014-12-10
Added league/event to the Event list.

frankdejonge:


sagikazarmark:
Can be exchanged with FuelPHP event. See #302

ziadoz:
Thanks, I'll get this merged. And thanks for all your hardwork on Flysystem too, I finally got the chance to use it before Christmas and it's really neat.

frankdejonge:
@ziadoz thank you for having it on the list :+1:

PR #284, opened 2014-12-15
Added Simply Random as random numbers generating alternative

rubensayshi:
Very simple library that only uses MCRYPT_DEV_URANDOM, which is 'the right thing to do' instead of all the libs that go and do their own custom mixing of various other libs.

see:
http://timoh6.github.io/2013/11/05/Secure-random-numbers-for-PHP-developers.html

and more:
https://news.ycombinator.com/item?id=6216101
http://security.stackexchange.com/questions/3936/is-a-rand-from-dev-urandom-secure-for-a-login-key/3939#3939


ziadoz:
Thanks for the suggestion.

PR #287, opened 2014-12-18
Add reference to Elephant.io

Taluu:
Another library I rewrote from scratch, used to send events (not yet receiving them) over the websocket protocol


ziadoz:
Thanks, I've merged this.

PR #291, opened 2014-12-27
Added JeyKeu/Notify

JeyKeu:
Added JeyKeu/Notify. A Bootstrap3 notifications for PHP


ziadoz:
Thanks for the suggestion.

PR #292, opened 2014-12-30
ArrayPath

mathiasgrimm:
Adding helper class


ziadoz:
Thanks for the suggestion.

mathiasgrimm:
Thanks!

Is is going to be merged into?

PR #293, opened 2014-12-31
Add `Magallanes` as deployment tool

kuldipem:


ziadoz:
Thanks for the suggestion.

BrunoViera:
Hello! we use a lot this tool in our company and side projects, would be great have it in the list.
Thanks in advance

PR #294, opened 2014-12-31
Change PHP Weekly Url

naxhh:
.info URL does not work anymore.

I've found this one: http://www.phpweekly.com/archive.html

And I think it's the same one.


t1gor:
This is the address that is present in the newsletter.

ziadoz:
Thanks for spotting this. I've merged the new URL.

PR #299, opened 2015-01-05
SmallMVC, a tiny&small framework.

Hacksign:
[SmallMVC](https://github.com/Hacksign/SmallMVC) framework is forked from tinymvc,but rewrite almost everything.
MVC framework.
Build in Router&Viewer, you can replace it with modifing config file in your project.
Convenient DB operations and display a template page.
[Demo Site](http://www.debugwars.com)


ziadoz:
Thanks for the suggestion.

PR #304, opened 2015-01-08
add CodeIgniter framework

CodeIgniter:
CodeIgniter is a powerful PHP framework with a very small footprint, built for developers who need a simple and elegant toolkit to create full-featured web applications.


ziadoz:
Thanks for the suggestion.

PR #307, opened 2015-01-12
Add Analog micro logging package

beheh:
[jbroadway/analog](https://github.com/jbroadway/analog) is a micro logging package by Johnny Broadway, small and simple like KLogger but with a static class and closures for maximum flexibility.


ziadoz:
Thanks, I've merged this.

PR #309, opened 2015-01-14
Add a mobile detect library

borishuai:
Mobile_Detect is a lightweight PHP class for detecting mobile devices (including tablets). It uses the User-Agent string combined with specific HTTP headers to detect the mobile environment.


ziadoz:
Thanks, I've merged this.

PR #316, opened 2015-01-18
Piwik/device-detector for universal device detection that can parse any user agent

nickurt:


ziadoz:
Thanks, I've merged this.

PR #318, opened 2015-01-18
Added Wjs lazy loader

weeger:
Added Wjs Library. Please, if rejected, let me some constructive comment.

http://github.com/weeger/wjs


ziadoz:
Thanks for the suggestion.

tleb:
This is a Javascript library, I believe.

> ziadoz/awesome-**php**

weeger:
This is both, an update is on the road with a more convincing demo

PR #323, opened 2015-01-21
Add HttpRefererFilter to HTTP section

GabrielJMJ:
Library URI: https://github.com/GabrielJMJ/HttpRefererFilter


ziadoz:
Thanks for the suggestion.

PR #324, opened 2015-01-23
Include talk about TDD and TPP

rezonanc:


ziadoz:
Thanks. Instead of adding a single video I've added the conference's channel. If anyone knows any other good PHP conference video channels let me know and I'll add them.

PR #331, opened 2015-02-04
Added Nameless debug package

corpsee:
Added Nameless debug package. https://github.com/corpsee/nameless-debug


ziadoz:
Thanks for the suggestion.

PR #332, opened 2015-02-04
Added PHPell

corpsee:
Added PHPell to Development Environment.  https://github.com/corpsee/phpell


ziadoz:
Thanks for the suggestion.

PR #334, opened 2015-02-07
add AFUP Youtube channel in the PHP Videos section

agallou:
it contains a lot of talks given during the PHPTour and ForumPHP conferences. Some of the talks are in english, but they are mostly in french.


omansour:
:+1:

ziadoz:
Thanks for suggestion this. I'd love to add these, but I can't really look through the videos as I don't speak any French unfortunately. Plus, at least for the time being, all the resources that I'll be adding will be in English. 

I have added a note to revisit adding resources in different languages to the enhancements ticket though, so they won't be forgotten, as it's something I'd love to do.

The same also applies to https://github.com/ziadoz/awesome-php/pull/335

PR #339, opened 2015-02-11
Add book: Head First Design Patterns

lluisi:


ziadoz:
Thanks, I've added this to the list.

PR #340, opened 2015-02-13
Add new section: Magazine, with a phparch link

lluisi:
I suggest to add a new section called Magazines, with the most popular: php[architect]


ziadoz:
Thanks, this looks great.

PR #343, opened 2015-02-15
Add ShouldPHP

GabrielJMJ:
For to test classes with a friendly syntax. https://github.com/GabrielJMJ/ShouldPHP


ziadoz:
Thanks for the suggestion.

PR #345, opened 2015-02-16
Add Stop

gherkins:
https://github.com/ivoba/stop

Stop is a lightweight Dumper for PHP 5.3+

* shortcut functions
* some more infos as File, Line & Memory
* some nicer rendering with Twitter Bootstrap,
* options for return or continue & hide which will print to javascript console
*disable Stop for prod env
* autoloading
* extensibility, if youd like to use jqueryUI f.e just subclass the Dumper and set the class
* context awareness: if you are in console it will render in text mode, if in ajax mode it will render in json mode
* dumper for json strings
* stop_type as a better version of var_dump(get_class($var));exit;


ziadoz:
Thanks for the suggestion.

PR #346, opened 2015-02-18
Update README.md

mistaguy:
Added xcrypto security library


ziadoz:
Thanks for the suggestion.

PR #347, opened 2015-02-19
Adding new collection library

italolelis:
A new abstraction for collections, we follow all PSR's codes, very fast implementations, based on native DoublyLinkedLists, Queues, Stacks, Trees and others. Looks like Ardent, but it's a little more robust because we're implementing Reactive Extension and LINQ into a unified library. Please take a look into the code and see what it can do.

Collections Abstraction library for PHP

The Collection library is one of the most useful things that many modern languages has, but for some reason PHP doesn't has a built in collection layer.

For that reason we created Collections, an incredible library that gathers the best of .NET's and Java's collections patterns and unify it with PHP array power.

Take a look and see what we're talking about!!


vladimino:
Hey, it's a great library! :+1:  
What do you guys think?

PR #349, opened 2015-02-26
Yii got 2 versions

vilkh3m:
Yii version 1 is very popular so i thing is good to be in list


SilverFire:
+1

ziadoz:
Thanks for the suggestion.

PR #350, opened 2015-02-27
php dependency injection lib added

mehmetsoylucom:


ziadoz:
Thanks, I've done this via: https://github.com/ziadoz/awesome-php/pull/440

PR #352, opened 2015-03-03
Add Writeln

GabrielJMJ:
Fast and simple Symfony Console command creation https://github.com/GabrielJMJ/Writeln


ziadoz:
Thanks for the suggestion.

PR #368, opened 2015-04-17
Add Lithium 3 framework

agborkowski:
add li3 framework


josegonzalez:
Once they have a stable 1.0.0 release, I would be happy to merge this.

PR #367, opened 2015-04-15
Added Lumen to Micro Frameworks

efik:
Hello!:)
I think this should be here.

PR #377, opened 2015-05-11
Added Lumen to the Micro Frameworks section

inphomercial:
This is an awesome resource you've helped put together, thanks!
I added the new Lumen micro framework to the list.


UFOMelkor:
Duplicate of #367

PR #383, opened 2015-05-19
Added Slim Eloquent

iGusev:
The Slim PHP micro framework paired with Laravel's Illuminate Database toolkit.


ziadoz:
Thanks for the suggestion.

PR #384, opened 2015-05-19
[php books] XML Parsing with PHP

iGusev:
This book covers parsing and validating XML documents, leveraging XPath expressions, and working with namespaces as well as how to create and modify XML files programmatically.


josegonzalez:
<3 that xml needs a whole book for developers to figure out...

iGusev:
@josegonzalez done!

PR #385, opened 2015-05-19
Added Creating PHP MVC Book

faizanayubi:
An Important Book every PHP Programmer should read


panique:
@faizanayubi Definitly an interesting book, but the link shows an illegal download page. I think it's not fair for the authors to link such sites within this project. Could you add link to the publisher's page please ?

josegonzalez:
Closing as it's not the best link...

PR #386, opened 2015-05-19
Update Omnipay to reflect The PHP League link.

kayladnls:


ziadoz:
Thanks for fixing this!

PR #389, opened 2015-05-20
Add Fakerino a modern fake data generator framework

niklongstone:


tleb:
I am not certain about the integration of this repo in this list:

* really recent (first commit 22/02/15);
* only one contributor;
* why is this installation through a `composer create-project` and not a simple `require` ? I don't think a fake data generator should be the base of an app.

I haven't tested the library, but I think that more active users is needed for a lib to go in this list.

niklongstone:
so, please update the CONTRIBUTING.md inserting minimum requirements for this list.

tleb:
I do not manage this list, I just give my opinion. @ziadoz might have a completely different opinion.

niklongstone:
I understand now why Justin Bieber is so popular....

tleb:
Euh, what?

* PHP
* Github
* list
* awesome
* contributing
* Justing Bieber

Find the link. You have two hours.

beautyface:
I'm to pretty to read this.

PR #399, opened 2015-05-28
[routers] baryshev/TreeRoute

baryshev:
https://github.com/baryshev/TreeRoute

This is a new performance focused PHP routing library.
This is a fastest PHP router right now.

Here is a benchmark suite: https://github.com/baryshev/php-router-benchmark
Another benchmark: https://github.com/baryshev/FastRoute-vs-TreeRoute


josegonzalez:

PR #400, opened 2015-05-30
Added punic/punic in internationalization and localization

JosielSantos:
Documentation: http://punic.github.io/


josegonzalez:
This has merge conflicts, so I won't be merging it.

PR #402, opened 2015-05-31
Add simple-flash

tamtamchik:
https://github.com/tamtamchik/simple-flash - Easy, framework agnostic flash notifications.


ziadoz:
Thanks for the suggestion.

PR #408, opened 2015-06-10
[code analysis] pecl: package/vld

patrickallaert:
https://pecl.php.net/package/vld


josegonzalez:

PR #410, opened 2015-06-10
Added PEAR/PECL repositories

patrickallaert:
https://pear.php.net/ + https://pecl.php.net/


josegonzalez:
Closing. Happy to discuss, but I think most developers see pecl/pear as old-hat, not awesome. There are certainly things there that *can* be on this list, but pushing developers to packages not available via composer etc. is I think not the point of this list.

PR #414, opened 2015-06-10
Reorganizing Debugging with Profiling

patrickallaert:
... and moving XHProf to that merge while fixing its name (using original case sensitivity)


ziadoz:
I think a standalone Profiling category might be better. If you tweak the PR to do that I'll get it merged. Thanks.

patrickallaert:
I considered that too at first, but given that it is quite common (e.g.: Xdebug, Tracy and Z-Ray) that a tool mixes both aspects, I wondered how to manage that correctly and haven't found a good solution.

What about copying the lines into the different categories? Not a very elegant solution, but it would do the trick.

ziadoz:
Good point. In that case I think your category name is probably a decent improvement. Thanks.

PR #417, opened 2015-06-16
added CodeIgniter as framework

hatemzidi:
hello, i suggest to add CodeIgniter too, it's a lightweight web framework.


tleb:
[Already suggested](https://github.com/ziadoz/awesome-php/pulls?q=codeigniter).

hatemzidi:
My mistake ... but i didn't see it in the list. so i just pull requested it (again)

jacobsantos:
I think having CodeIgniter on the list would diminish the list just a little bit. Fuel might be a good addition, if it isn't already on the list, but I think Laravel is a better choice than FuelPHP. Don't get me wrong, FuelPHP is some good code. I would put that code in a pipe and smoke it. I just think Laravel does a lot of the things Fuel was doing better and easier.

I worked with CodeIgniter 1 and 2 and the code is no where near awesome. I took a look at version 3 and I think it is just 2 with a better or at least more permissive license.

PR #428, opened 2015-07-17
Change Eloquent text, its not just specific to Laravel 4

kabirbaidhya:


ziadoz:
Thanks. I've amended things to make it clearer.

PR #429, opened 2015-07-19
Update url of zend framework components

UFOMelkor:
The locations of the zend framework components have changed. Old links lead to a 404 page.


ziadoz:
Good spot. Thanks!

PR #430, opened 2015-07-19
added MysqliDb

avbdr:
Very simple to learn, easy to use mysql query builder and mini Orm.


ziadoz:
Thanks for the suggestion.

PR #431, opened 2015-07-26
[data structures and storage] Raphhh/balloon

Raphhh:
Here is a suggestion to include Balloon, a tiny file data access layer:
https://github.com/Raphhh/balloon


josegonzalez:
While this looks like a great project, for a library that handles file data access, I think it's a bit young to add to a list of awesome tools. Looking forward to seeing how it turns out!

Raphhh:
ok, thank for your time :)

PR #440, opened 2015-08-03
Fixed the PHP-DI entry

mnapoli:
Hi there, thank you for mentioning PHP-DI. I have updated the entry:

- the name is PHP-DI (instead of PHP DI)
- updated the URL (http://php-di.org/)
- updated the description (annotations have become a secondary feature and are encompassed by "autowiring")

Please let me know if there's anything to change. If you want I can replace the description with a more generic "Another dependency injection container", it would still be better than mentioning annotations.


ziadoz:
Cheers, I've merged this.

PR #443, opened 2015-08-05
Clarify CLImate's description

klaude:
I love this list. :) CLImate does more than CLI output, so I briefly described its other functions.


ziadoz:
Thanks, I've merged this.

PR #447, opened 2015-08-06
[dependency management extras] joshdifabio/composed

joshdifabio:
[Composed](https://github.com/joshdifabio/composed) is a bit like [Composition](https://github.com/bamarni/composition) but solves a slightly different use case.


josegonzalez:
Can you rebase this PR?

joshdifabio:
Done.

PR #450, opened 2015-08-10
Add FuelPHP

xrobau:
More info http://fuelphp.com/features


tleb:
[Already suggested, already explained](https://github.com/ziadoz/awesome-php/pull/85#issuecomment-41633053).

xrobau:
That links to a pull request saying no to Cake, but Cake is now there?

tleb:
> Guidelines explicitly say: "Libraries that are PHP 5.3+, Composer-installable, PSR compliant, tested and documented are preferred.". I see a lack of PSR in both and a lack of tests in cake

> Thanks for the suggestions. Both of these have been suggested before and I'm also happy with the current selection of frameworks on the list.

dereuromark:
Cake is now PSR,
but IMO PSR is not about functionality, and thus should NOT be criterium for acceptance!
It should not matter at all what coding standards, as long as it is composer-inst and tested.
PSR is a guideline and "optin", but not a must have.

josegonzalez:
Not PSR.

PR #458, opened 2015-08-21
Add HTML to Markdown

colinodell:


colinodell:
The package may seem young (with only 830 downloads), but it was migrated directly from  nickcernis/html-to-markdown (with 32,811 installs) into the League's namespace and cleaned up a bit.

colinodell:
As of today this project has 69,881 downloads (22,085 under the new namespace, 47,796 under the old one).  What are your thoughts on including this in your list?

PR #462, opened 2015-08-28
[php books] The Clean Architecture in PHP

brpaz:
I read this book (https://leanpub.com/cleanphp) and I found it awesome. Its very well written. I think its a must read for any PHP developer.


josegonzalez:

PR #465, opened 2015-09-01
[exercises] danrevah/php-exercises

danrevah:
GitHub Repository: https://github.com/danrevah/php-exercises

Based on the same idea of the kolodny javascript exercises (https://github.com/kolodny/exercises).
using the TDD approach to practice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 	- [PHP Videos](#php-videos)
 	- [PHP Reading](#php-reading)
 	- [PHP Internals Reading](#php-internals-reading)
+- [Exercises](#exercises)
+	- [PHP Exercises](#php-exercises)
 - [Contributing](#contributing)
 
 ## Dependency Management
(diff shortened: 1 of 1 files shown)


tleb:
@danrevah: If you give exercises to beginners, I think you should give a good example with your code. In your `test.php`, you declare new symbols and you have logic. This goes against most standards.

Anyway, I like the test-unit-verified exercises idea.

danrevah:
@tleb Thank you for your feedback, but in each exercise directory there's a readme file that explains exactly what needs to be done.
I think it's pretty straightforward.  Anyway if you have any ideas on how to improve it, you are more than welcome to suggest pull request.

tleb:
What I am saying is that if a beginner looks at your code (curiousity can be an advantage), they will see logic and symbol declarations in the same file, which is bad a bad practice.

danrevah:
I don't agree with you. It's following the same idea behind the kolodny js exercises which are aimed to beginners as well.

josegonzalez:

danrevah:
@josegonzalez https://github.com/ziadoz/awesome-php/pull/637

PR #472, opened 2015-09-19
Add Climb

duxet:
https://github.com/vinkla/climb


tleb:
> This tool is still in early stages of development and may change in the future.

Shouldn't packages in this list be stable?

One thing: I am not sure about what this achieves? Is it just a `composer update` alias? The readme isn't really helping.

duxet:
It depends on what you mean by stable - it's working right now, so i think it shall pass.

`composer update` only updates dependencies to version requirement set in composer.json. This tool works like [ncu](https://github.com/tjunnone/npm-check-updates) [bcu](https://github.com/se-panfilov/bower-check-updates) and shows also updates, that don't match the version constraint.

shadowhand:
Already in the list. This should be closed.

PR #477, opened 2015-09-30
[routers] bephp/router

lloydzhou:
add Router library. A barebones router for PHP. It matches urls and executes PHP functions, and automatic get variable based on handler function parameter list.


lloydzhou:
there's one router benchmark result here:
https://github.com/tyler-sommer/php-router-benchmark/issues/11#issuecomment-149795110

"bephp/router" is faster than "Pux" and "FastRoute".

josegonzalez:

PR #482, opened 2015-10-06
Remove webadvent - link is not working

peterkokot:


josegonzalez:
Already removed. Thanks!

PR #484, opened 2015-10-08
[arrays] bocharsky-bw/Arrayzy

bocharsky-bw:
Propose to add [Arrayzy](https://github.com/bocharsky-bw/Arrayzy) to the list.

This library is:
* The wrapper for all PHP built-in array functions and easy, object-oriented array manipulation library. In short: Arrays on steroids;
* PHP 5.4+;
* Composer-installable;
* PSR-4 compliant;
* 98% tested with PHPUnit;
* well documented;

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
     - [Command Line](#command-line)
     - [Authentication and Authorization](#authentication-and-authorization)
     - [Markup](#markup)
+    - [Arrays](#arrays)
     - [Strings](#strings)
     - [Numbers](#numbers)
     - [Filtering and Validation](#filtering-and-validation)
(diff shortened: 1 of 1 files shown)


ecolinet:
:+1:

josegonzalez:
There are other libraries that handle this which are [already listed](https://github.com/ziadoz/awesome-php/blob/master/README.md#data-structure-and-storage), so without a good reason, I'll abstain from merging for now. Regardless, thanks for your contribution.

bocharsky-bw:
@josegonzalez Actually, this is not another library of collection implementation / data structure implementation. This is something different.

Arrayzy is a wrapper around all PHP built-in array functions which allows developers to work with native PHP arrays in an object-oriented way. I don't see any similar library in the list.

I'm not sure about creating a new section for work with built-in arrays only, but I still insist on merging this suggested library to `ziadoz/awesome-php`. What do you think?

josegonzalez:
Any/all of the existing collections libraries have significant overlap with your library.

PR #491, opened 2015-10-20
[continuous integration] wercker.com

fruitl00p:
Wercker is an easy to use CI platform (like Travis) but more suited for also deploying those apps directly into production for quick and painless (i.e.) microservice deployments.


josegonzalez:
Manually merged in 25868b74fa4dd2cf2898a61e53ee167a23a6ec30.

PR #493, opened 2015-10-23
[http] 42mate/PHRequests

jonafrank:
I'm using this library for a while and it's been useful to do simple HTTP requests. It uses cURL to make the requests and is easy to consume REST resources.

https://github.com/42mate/PHRequests


josegonzalez:
There are other libraries that handle web requests which are already listed, so without a good reason, I'll abstain from merging for now. Regardless, thanks for your contribution.

PR #495, opened 2015-10-24
Sort list of resources alphabetically

henriquemoody:


henriquemoody:
This pull request got old and now it's impossible to merge without a rebase.

I will be glad to do it since I got sure that it is going to be merged.

The problem I see is that, if this pull request got merged, the other 100 pull requests will probably need to be updated/rebased.

What should I do @ziadoz?

josegonzalez:
I'm going to close. I'll go through all the PRs soon and get them merged, at which point we can revisit this.

PR #496, opened 2015-10-25
Suggest PHPSA

ovr:


ovr:
ping @ziadoz

jkufner:
+1

This looks very nice! Thank you.

jadb:
I see the library is still in alpha. I think it's awesome to have such a library but maybe wait until it's out of alpha at least? 

Closing for now but would definitely reconsider when out of alpha.

PR #497, opened 2015-10-25
[templating] smarty.net

guilhermesiani:


josegonzalez:
Thanks, but we don't need smarty twice :P

PR #498, opened 2015-10-26
added http://htmlpurifier.org/ to Filter/Validation

jakoch:


README.md
```diff
@@ -582,6 +582,7 @@ Libraries to help manage database schemas and migrations.
 ## Filtering and Validation
 *Libraries for filtering and validating data.*
 
+* [HTMLPurifier](http://htmlpurifier.org/) - A HTML filtering library.
```
bocharsky-bw: Maybe we should point it to the [GitHub](https://github.com/ezyang/htmlpurifier) repo?

tleb: @bocharsky-bw Not sure about this one. The GH doesn't provide any information, apart for the site url, which is where everything is needed to begin with the lib.

bocharsky-bw: GitHub repo provide source code at first, `composer.json` for install via composer that's defacto PHP standart library installation for now, and the link to the official docs website. I vote :+1: for GitHub link

ziadoz:
This is already on the list. Thanks.

PR #499, opened 2015-10-27
[event] icicleio/icicle

coderstephen:
Adds [Icicle](https://github.com/icicleio) to the "Event" section. Icicle is an awesome asynchronous library that has packages for non-blocking I/O, socket servers, and even true multithreading and worker pools!

The core library is [`icicleio/icicle`](https://github.com/icicleio/icicle), which provides event loops, promises, and coroutines to make writing asynchroous code much easier to understand without descending into callback hell.


coderstephen:
Bump?

PR #501, opened 2015-10-30
[office] nuovo/spreadsheet-reader

justinvoelkel:
Nice package for extracting data from a few different types of spreadsheets - https://github.com/nuovo/spreadsheet-reader


josegonzalez:
There are other libraries that handle this which are already listed, so without a good reason, I'll abstain from merging for now. Regardless, thanks for your contribution.

PR #511, opened 2015-11-11
[api] skrz/meta

jakubkulhan:
A library to map from PHP arrays, JSON, XML and Protocol Buffers (protobuf) to PHP objects and back.

https://github.com/skrz/meta


josegonzalez:
I personally think this is a solution in search of a problem. Thanks for the pr.

PR #512, opened 2015-11-12
[database] kassko/data-mapper

kassko:
https://github.com/kassko/data-mapper


josegonzalez:
While this looks like a great project, for a library that interfaces with databases, I think it's a bit young to add to a list of awesome tools. Looking forward to seeing how it turns out!

PR #520, opened 2015-11-23
Added a Environment library

jpcercal:
Added a Environment library that can be used to get (simple and complex) values and convert it to php data types using the environment variables.


josegonzalez:
Sorry, it's not a very popular package, and there are already similar packages out there that are commonly used.

PR #521, opened 2015-11-25
[templating] FoilPHP/Foil

KazeFlame:
A good alternative to Plates native template engine. http://www.foilphp.it/


bocharsky-bw:
I think better to add link to the GitHub repo instead.

KazeFlame:
Done :smile:

PR #522, opened 2015-11-26
[internationalisation and localisation] marxo/php-zlo

marxo:
I've added php-zlo as a library for inline translation in this PR.

https://github.com/marxo/php-zlo


josegonzalez:
While this looks like a great project, for a library that does internationali(sz)ation, I think it's a bit young to add to a list of awesome tools. Looking forward to seeing how it turns out!

PR #523, opened 2015-11-29
[authentication and authorization] knpuniversity/KnpUGuard

bocharsky-bw:
The [KnpUGuard](https://github.com/knpuniversity/KnpUGuard) adds simple and beautiful authentication to Symfony's security component in Silex and anywhere else.


josegonzalez:
> This library is deprecated since Symfony 2.8 and won't work with Symfony 3.

PR #524, opened 2015-12-02
[templating] rybakit/twig-extensions-deferred

rybakit:
https://github.com/rybakit/twig-extensions-deferred


josegonzalez:
Seems like twig should be handling their own "best of" plugins, given that there are so many.

PR #525, opened 2015-12-03
[architectural] antares993/Accessible

antares993:
[Accessible](https://github.com/antares993/Accessible) is a PHP library that allows you to define your class' getters, setters and constructor in an elegant and powerful way using docblock annotations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,7 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 * [Functional PHP](https://github.com/lstrojny/functional-php) - A functional programming library.
 * [Lib Accessor](https://github.com/phine/lib-accessor) - A library for simplifying accessors.
 * [Iter](https://github.com/nikic/iter) - A library that provides iteration primitives using generators.
+* [Accessible](https://github.com/antares993/Accessible) - Define your class' getters, setters and constructor using powerful annotations.
 
 ## Debugging and Profiling
 *Libraries and tools for debugging and profiling code.*


jkufner:
Why would anyone want such thing?

antares993:
Tipically, you may want to use it in domain classes or DTOs. Most of the time, these kinds of classes have a set of attributes with their getters / setters, and simple rules are defined inside the setters to disallow the user to set a wrong value to an attribute. These classes have a simple logic, but also have a big size and are hard to read compared to what you can have in C# or Scala for example.

The objective of this library is to have these kinds of classes easier to read and understand. If you want to analyze a class, instead of searching the getters / setters of each attribute to know if they exist and what kind of value they accept, you just have to look at the docblock of the attribute: a one-line definition of its access is provided, with eventually a set of Symfony constraints. This way, the size of your class will be dramatically reduced, and the only methods you will find in the class will concern non-trivial logic.

For example, for an attribute `$email` that cannot be set to `null` and can only accept strings that match the email format, you would simply have something like this:

```php
/**
 * @Access({Access::GET, Access::SET})
 * @Assert\NotNull
 * @Assert\Email
 */
private $email;
```

keradus:
:-1: for adding not-known and not-stable and not-anything projects
1watch + 5 stars...

antares993:
This library has stable versions, but that's still a good point. Maybe adding a criteria in `CONTRIBUTING.md` such as "The library must have a stable version and 50+ stars" would be a good idea.

keradus:
stable in not semver meaning but community/usage/world spread ;)

idea behind library is really nice and I like annotation-driven programming, but still lib is far from getting widely used

josegonzalez:
Do you mind rebasing this onto master?

josegonzalez:
Note: I think most the architectural libraries are pretty bullshit, so this isn't that far off from other stuff in that section, and is basically how you write code in certain orms so...

PR #531, opened 2015-12-09
Added `check_php` nagios plugin

cytopia:
A nagios plugin that checks PHP setups against the following:

* Check for PHP startup errors
* Check for PHP updates (inside minor versions and patch levels)
* Check for missing PHP modules
* Check for blacklisted PHP modules and throw err/warn if they are compiled in
* Check for expected php.ini config directives (e.g.: date.timezone must be "Europe/Berlin", etc)
* Each check can specify its own severity (warning or error)


ziadoz:
Thanks for the suggestion.

PR #533, opened 2015-12-20
Add Spress

yosymfony:
Spress is a static site generator which allows you to write and publish your web sites.


lury:
+1

This one is actually actively developed - not like the other two.

jadb:
Thanks!

PR #534, opened 2015-12-24
Update README.md

Lao-liu:
Add hprose


tleb:
That project seems interesting. :)

Shouldn't the link point to the PHP repo? There is a list of the languages in the README, so it seems fine to me.

PR #543, opened 2016-01-15
[templating] Talesoft/tale-jade

TorbenKoehn:
Tale Jade is a powerful implementation of the Jade template language for
PHP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 * [PHPTAL](http://phptal.org/) - A PHP implementation of the [TAL](http://en.wikipedia.org/wiki/Template_Attribute_Language) templating language.
 * [Plates](http://platesphp.com/) - A native PHP templating library.
 * [Lex](https://github.com/pyrocms/lex) - A lightweight template parser.
+* [Tale Jade](https://github.com/Talesoft/tale-jade) - A PHP implementation of the Jade template language.
 
 ## Static Site Generators
 *Tools for pre-processing content to generate web pages.*


josegonzalez:
There are other libraries that handle templating which are already listed, so without a good reason, I'll abstain from merging for now. Regardless, thanks for your contribution.

TorbenKoehn:
Well, you're not wrong, but Jade is the only template engine that exists that is **actually** different from all other ones. Jade is not your usual `{{foo}}`-engine.

It's more an own markup language. It acts on indentation rather than closing tags, you can write the same amount of HTML with just a fraction of the characters you'd need with plain HTML.

You're not adding Tale Jade, because there are enough Template Engines, but look at the existing ones, most of them are basically **Twig** with two or three different features.

Tale Jade is the only exceptional one.

If you want to avoid having too many template engines, at least provide diversity. Remove the **Twig**-like engines and add Tale Jade.

Also, it's not that Jade doesn't already exist, it's a well known JavaScript template engine and this one is not just a fork or port, but rather a completely specialized implementation for PHP.

I suggest you think about it again.

jadb:
@TorbenKoehn, please submit as new PR. I can see it being reconsidered for inclusion as it is the only one on the list for Tale Jade support, so could be helpful to anyone looking for that.

TorbenKoehn:
Hello,

thank you for reconsidering this.

The new PR can be found at #624 

The name of the template language is `Jade` and it's really popular in the Node.js world, the implementation is called `Tale Jade`. It has a bunch of bonus-features, but it's primarily the only complete, stable and fully working implementation of Jade in PHP.

Filisko:
Thank you @TorbenKoehn, I did not know it but it looks really nice.

PR #554, opened 2016-01-28
Add library for validation - Simple PHP request validator

progsmile:
Adding new library "Simple PHP request validator" - it's unique, documented, tested and very convenient tool for validation) 
It's "young" library, I think it will become very popular because of simplicity)


josegonzalez:
Doesn't pass this test:

> b) Well known or discussed within the PHP community

Also, it's functionality is well-provided by the existing documented options. I'm happy to revisit if/when this gains traction.

PR #556, opened 2016-01-29
Added Slimdump

dritter:
Slimdump is a small dumper tool for MySQL.


jadb:
@dritter nice tool but probably listed in the wrong section. 

Please move it to "Miscellaneous" as it is not an ORM package per se like the other ones listed under "Database".

dritter:
@jadb Done. :)

jadb:
Thanks!

PR #561, opened 2016-01-30
Add phpRedisAdmin

rezonanc:
Simple web interface to manage Redis databases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -768,6 +768,7 @@ Libraries to help manage database schemas and migrations.
 * [PHP Queue](https://github.com/CoderKungfu/php-queue) - An application for managing queueing backends.
 * [MailCatcher](https://github.com/sj26/mailcatcher) - A web tool for capturing and viewing emails.
 * [Cachet](https://github.com/cachethq/cachet) - The open source status page system.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -768,6 +768,7 @@ Libraries to help manage database schemas and migrations.
 * [PHP Queue](https://github.com/CoderKungfu/php-queue) - An application for managing queueing backends.
 * [MailCatcher](https://github.com/sj26/mailcatcher) - A web tool for capturing and viewing emails.
 * [Cachet](https://github.com/cachethq/cachet) - The open source status page system.
+* [phpRedisAdmin](https://github.com/ErikDubbelboer/phpRedisAdmin) - A simple web interface to manage [Redis](http://redis.io/) databases.
```
jadb: Maybe move to miscellaneous?

rezonanc: I was under the impression that all stand-alone web apps should go under 'Software'/'Web applications'. If it would be a library then I would agree.

Maybe I did not understood correctly? If that's the case, then sure, we can move it to miscellaneous.

jadb: Hmm, you are right! I mis-read.

PR #576, opened 2016-02-08
Updated PINQ's description.

saulotoledo:
An easier to understand description. It explains what is LINQ. Also, LINQ should be in uppercase.


jadb:
Thanks!

PR #594, opened 2016-03-06
Added ErrorDumper

bkrukowski:


jadb:
Thanks. Might include it when it gets more popular.

PR #610, opened 2016-03-22
[security] pachico/magoo

pachico:
Suggested Magoo package (mask credit card numbers, emails and more).


josegonzalez:

PR #611, opened 2016-03-25
Suggest Manipulator package

Nayjest:
It's a small library for manipulating PHP objects: https://github.com/Nayjest/manipulator


josegonzalez:
The library itself seems interesting, but is not widely used, so doesn't meet our criteria for addition. Thanks for the PR though.

PR #621, opened 2016-04-07
[build tools] jonathantorres/construct

raphaelstolt:
Added `construct` to the `build tools` section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,6 +412,7 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 
 * [Bob](https://github.com/CHH/bob) - A simple project automation tool.
 * [Box](https://github.com/box-project/box2) - A utility to build PHAR files.
+* [Construct](https://github.com/jonathantorres/construct) - A PHP project/micro-package generator.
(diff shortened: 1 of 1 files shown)


raphaelstolt:
What about that one? `Merge` or `close` please.

josegonzalez:
Closing. Composer actually has this functionality via `create-project`, which seems like the more flexible way of doing things.

raphaelstolt:
That's like comparing 🍎's to 🍒, but ✌️ .

josegonzalez:
Well I'm willing to discuss. How is this project better? It didn't seem like it was breaking new ground.

josegonzalez:
Right. I think what I'm saying is that you can create a custom project repository and use that as a skeleton for whatever you desire. All major frameworks have such a project, and composer users are more than encouraged to create their own.

raphaelstolt:
Sure you can use a skeleton, but does it follow established conventions and in how many spots do I have to adjust things (e.g. package name)? 

Which frameworks allow the creation of a framework agnostic skeleton project? Really not aware of one. 

And your last sentence indicates an intrinsically solution which once again might turn away from established conventions.

PR #632, opened 2016-04-20
[Authentication and Authorization] Add tymondesigns/jwt-auth

otezz:
JSON Web Token Authentication for Laravel & Lumen


josegonzalez:
Going to close this. We have a json web token library that doesn't specifically market itself as a laravel/lumen library. Please re-submit this to whatever the awesome-laravel list is, as it'll be much more useful there.

PR #636, opened 2016-04-21
[static site generators] jigsaw

raphaelstolt:
Added Jigsaw to the static site generators section.


josegonzalez:
I haven't merged this because I'm trying to think of a good metric for "awesome" repositories. Not to say your code isn't great, just don't want this list to become a "here is all the php code!" list. 

Will almost certainly involve removing some codebases where we have multiple implementations of the same thing.

raphaelstolt:
It's not __my__ code, but IMHO it's defo a static site generator worthy to mention and list.

jadb:
No tests 😞

raphaelstolt:
Ok that's a valid metric point. :smile:

PR #639, opened 2016-04-22
[Templating] Add rotexsoft/file-renderer

rotexdegba:
Added **[PHP File Renderer](https://github.com/rotexsoft/file-renderer)** to the Templating section.


ziadoz:
Thanks for the suggestion.

PR #640, opened 2016-04-23
[Data Structure and Storage] Add msgpack.php

rybakit:
https://github.com/rybakit/msgpack.php


josegonzalez:
I like it. Are there other pure php alternatives? Cursory search says no, just wanted to check though.

rybakit:
I'm not aware of any _working_ implementation of the MessagePack protocol in pure php. Actually, this library was born because I didn't find anything suitable ;)

rybakit:
@josegonzalez ping

ziadoz:
Thanks, I've merged this.

PR #643, opened 2016-04-28
[testing] Add PHP-Mock

malkusch:
PHP-Mock is a library to mock built-in PHP functions (e.g. time()).

See also: https://github.com/php-mock/php-mock


tommy-muehle:
👍

PR #646, opened 2016-05-05
add ThinkPHP

yunheli:


tleb:
* Links to 404.
* Description should probably be in english.

yunheli:
@tleb  I am so sorry

tleb:
The sad thing is that this lib has no english documentation, so those who find it in this list will not be able to use it.

yunheli:
I see, thank you @tleb

tleb:
I am no maintainer, so I have no word on what happens with this, just my opinion.

Is there no list in chinese you could add this to?

Jaggle:
@tleb most Chinese use thinkPHP, but it is really not good designed

PR #647, opened 2016-05-06
Update README.md

ch3k1:


jadb:
Sorry, but not popular enough and no tests I could find. Please re-submit when ready.

PR #648, opened 2016-05-07
Add PHPT to Tests section

jkufner:


tleb:
I'm not sure it has a place in this list, isn't it to test PHP's core? It isn't a library, right?

jkufner:
It is a tool, not library. It is used to test PHP core (which alone makes it a proved tool), but it can be easily used for testing any PHP code. Its beauty is in simplicity and being easy-to-use, but it is very powerful tool.

tleb:
Ok, it can be used to test any PHP code. The main thing is: is it actually used by any lib/framework? Is it even its recommended use, to test any PHP code, not only PHP's core? I haven't seen anything which recommends using it on everyday code in the link you added.

jkufner:
Here: https://github.com/search?utf8=%E2%9C%93&q=extension%3Aphpt+%22--file--%22&type=Code&ref=searchresults

> We’ve found 3,361,905 code results

tleb:
I don't think that's valid.

A few thousands are just useless ones with Foobar in them. Another few thousands are from libs which use it without composer. And a big part is PHP source and its forks.

I just want to know if you have an example of a *good* PHP lib which uses it just to make sure it checks the first two tests to enter this list:

1. Be widely recommended regardless of personal opinion
2. Well known or discussed within the PHP community

jkufner:
One good example: https://github.com/fpdo/fluentpdo/tree/master/tests (good tests, not so good library)

PHPT has its own tag on StackOverflow: http://stackoverflow.com/questions/tagged/phpt

Chapter in Pear manual: http://pear.php.net/manual/en/pyrus.commands.runphpt.php

A random blogpost: https://mwop.net/blog/64-PHP-Unit-Tests-and-the-winner-is-phpt.html

PR #649, opened 2016-05-09
[Filtering and Validation] Add Wixel/GUMP

filisko:


filisko:
@jadb hi?

josegonzalez:
How is this different/better than existing solutions we list?

filisko:
How different are the existing solutions from each other?

josegonzalez:
I don't know, but its on you to prove that your submission belongs on this list.

filisko:
I'm not the owner of this library, just a collaborator and I think that people may be interested in this library as is very easy to set rules and validate.

josegonzalez:
Right, but how is it different/better than the solutions we've listed? I'm sure the other libraries also are easy to use for adding rules and validation, otherwise they wouldn't be here.

filisko:
It has filtering integrated as well, so you can filter and validate.

PR #652, opened 2016-05-11
[PHP Installation] Add Brew PHP Switcher

jk2K:


jadb:
Thanks for the submission.

Is this similar to https://github.com/phpbrew/phpbrew? How does it differentiate?

jk2K:
`Brew PHP switcher` can switch Homebrew's PHP module location automatically in the Apache config, make full use of `Homebrew`, more easy

`Brew PHP switcher` can install php extension with `Homebrew`, while `PHPBrew` install php extension by itself

jadb:
Thanks for the update. I will let @josegonzalez look at this, he is a homebrew expert.

ziadoz:
Thanks for the suggestion.

PR #655, opened 2016-05-19
[ci] update white list

ReadmeCritic:


ReadmeCritic:
Hm looks like temporary issues
Closing

PR #659, opened 2016-06-04
[Code Analysis] Added phan

zakharovvi:
[phan](https://github.com/etsy/phan) - A static analyzer based on PHP 7+ and the php-ast extension.


ziadoz:
Thanks.

PR #660, opened 2016-06-05
Add lock

malkusch:
Lock is a lock library to provide exclusive execution of otherwise
concurrent code.

See also: https://github.com/malkusch/lock


josegonzalez:
Are there other libraries in this space?

malkusch:
I'm not aware of another library which provides synchronization through different backends.

If you search on packagist you'd find some implementations for a specific backend, e.g. this  https://github.com/ronnylt/redlock-php which implements only "Redlock" and only for ext-redis or search for "semaphore" to find specific Semaphore implementations. I failed to find something similar to the [TransactionalMutex](https://github.com/malkusch/lock#transactionalmutex), but that's maybe because users are not so aware that their code [should be prepared for failing transactions](https://dev.mysql.com/doc/refman/5.7/en/innodb-deadlocks-handling.html).

ziadoz:
Thanks, I've added this.

PR #664, opened 2016-06-12
added antidote to the migration tool list

dvwallin:


josegonzalez:
While the idea is nice, there are other, more popular solutions to this problem already listed here. I encourage you to continue working on this project and submit it when it better follows [our guidelines around contributing projects](https://github.com/ziadoz/awesome-php/blob/master/CONTRIBUTING.md) to the list :)

PR #665, opened 2016-06-14
Add 'Awesome PHP @LibHunt' to the 'PHP Websites' section

StanBright:
Awesome PHP @LibHunt is based on the awesome-php list here plus some helpful addition. For example, most libraries have popularity and activity index that helps you judge what to try, and you can also compare different libs. Moreover, I believe, it has a decent UI/UX that is easy to navigate/use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -846,6 +846,7 @@ Various resources, such as books, websites and articles, for improving your PHP 
 ## PHP Websites
 *Useful PHP-related websites.*
 
+* [Awesome PHP @LibHunt](https://php.libhunt.com/) - Your go-to PHP Toolbox.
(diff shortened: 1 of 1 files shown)


josegonzalez:
It's literally this list in a different format. 👎 unless someone else with commit bit wants to merge it.

ziadoz:
I think we'll pass on this.

StanBright:
@ziadoz fair enough.

I'm considering adding it to "Resources > PHP Websites (Useful PHP-related websites.)". I believe it could be counted as useful. Also, tracking the number of downloads (Packagist) is on its way. That metadata/info will be useful, too. What do you think? Can I open a PR?

PR #666, opened 2016-06-14
[Notifications] Minishlink/web-push

Minishlink:
Added the WebPush library that sends push notifications to browsers that support the [Push API](http://www.w3.org/TR/push-api/).


josegonzalez:
How is this different from any of the existing listed notification packages?

Minishlink:
The other listed notification packages are dealing with proprietary push notifications services (APNS, GCM, "Pushwoosh") or simple notification softwares for development (Growl, toaster, notify-send).

WebPush sends push notifications to standardized servers, which follow the [Web Push protocol](https://tools.ietf.org/html/draft-thomson-webpush-protocol-00). These servers are owned by browser vendors.

PR #669, opened 2016-07-02
Library officially abandoned - they recommend moneyphp/money

lukeholder:
> This project has been abandoned. It was only ever intended to be used as an example for PHPUnit features etc. and not for usage in production. I am sorry that I failed to make that clear. Please have a look at moneyphp/money instead.


ziadoz:
Thanks.

PR #673, opened 2016-07-14
[frameworks] add CodeIgniter

arjamizo:
add CodeIgniter Framework https://www.codeigniter.com/


josegonzalez:
Closing. CodeIgniter still supports [unsupported versions of php](https://github.com/bcit-ci/CodeIgniter/blob/3.0-stable/composer.json#L14), and I don't think that's particularly awesome for advancing the state of php. Happy to look at this again when that changes.

josegonzalez:
Okay? Please don't comment on a 2 year old issue, file a new one if you think this is a worthy inclusion.

PR #676, opened 2016-07-15
Added Stringizer, another string manipulation library

jasonlam604:
Kindly adding another String resource library


josegonzalez:
Why is this library better or worse than any of the existing libraries?

jasonlam604:
Well to be honest, I don't think it's necessarily better or worse, really subjective.  It is quite similar to Stringy but it does differ on a few features. But more or less the same type of library.  Is your intent to only have 1 library listed for each type of "area"?

josegonzalez:
The intent is to have the best library. There are quite a few metrics to judge that on - test coverage, feature-set, popularity, speed - and on the face of it, if there is a more popular library covering 90% of the feature set of a less popular library, I'd rather list the former than the latter (with exceptions of course!).

jasonlam604:
Ok fair enough, thank you for explanation.

PR #678, opened 2016-07-21
http://www.exakat.io has been down for more than a week.

defenestrator:
I don't know if it is coming back, but it has not been responding for some time now. Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,6 @@ rvm:
 before_script:
   - gem install awesome_bot
 script:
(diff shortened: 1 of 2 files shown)


defenestrator:
I probably should have squashed those commits, sorry. I can make a new PR if needed.

josegonzalez:
Just use the github link: https://github.com/exakat/exakat

josegonzalez:
Mind filing a new pr with the correct link instead?

PR #682, opened 2016-08-01
Add Windwalker framework

asika32764:


josegonzalez:
How is this different/better/worse than the existing frameworks that we list? As much as I am for great new stuff, I don't want to add people's pet projects - hence why I don't list any of my own side stuff.

PR #683, opened 2016-08-07
Add GImage

joseluisq:
https://github.com/joseluisq/gimage


josegonzalez:
How is this different/better/worse than the existing frameworks that we list? As much as I am for great new stuff - I think your library is pretty cool, I always hated dealing with image manipulation - I don't want to add people's pet projects without good reasons :)

joseluisq:
Really, I sent not this PR catalogued as "pet project", but by collaboration-related purposes.
Anyway, I understand your position about good reasons.

josegonzalez:
Cool, thanks for understanding :)

PR #691, opened 2016-09-05
FPDF

SusanthCom:
FPDF is a PHP class which allows to generate PDF files with pure PHP


jadb:
No clear license, not sure where on Github to find it to see how up to date it is without downloading it.

Also, if you resubmit, please don't include at top of the list 😉

SusanthCom:
Thanks. I will not put in top.

 https://github.com/Setasign/FPDF
There is no official git hub page. 

License, It's in public domain. 

I have used it extensively. Updates are very slow. 
But library is very stable and reliable for production Usage. 

Please include it at least in bottom of PDF section.

PR #698, opened 2016-09-25
[+]: add "Portable UTF-8"

voku:
One more string manipulation library, with UTF-8 safe replacement methods.

-> https://github.com/voku/portable-utf8#class-methods

---

This change is [Reviewable](https://reviewable.io/reviews/ziadoz/awesome-php/698)


josegonzalez:
Any reason this is better/different from the existing one?

voku:
I have bundled many string functions into one simple class via static methods + added missing tests, fixes, documentation + performance improvements, so you can simply switch from an non UTF-8 safe function. Compatible and Tested with php >= 5.3 - 7.1 + HHVM. 

Example: 
```php
var_dump(str_split('déjà', 2)); // https://3v4l.org/K78c8
var_dump(UTF8::str_split('déjà', 2)); // https://github.com/voku/portable-utf8/blob/master/tests/Utf8StrSplitTest.php
```

josegonzalez:
There is already another library we list right above yours. How is *your* library better than that one?

voku:
Nicolas Grekas has merged all logic from "Patchwork-UTF8" into [symfony/polyfill](https://github.com/symfony/polyfill/commits/master) + fixes. So I try to use the installed extensions if possible and as fallback we use symfony/polyfill.   

Example: UTF8::substr()
-> 1. if "mbstring" is really installed we use it
-> 2. if "iconv" is available we use that
-> 3. otherwise we use a workaround or a polyfill

josegonzalez:
So then does this new library supersede the patchwork library?

voku:
Theoretically, this could be replaced, but there are other parameters, namespaces, class-name etc., so to much BC. And I still use "Patchwork-UTF8" for some projects, some other projects only need "symfony/polyfill" and in the most cases you can use php's native functions. It's only one more option for people who need some special UTF-8 functionality like `````

voku:
3 years later :) ... and I am still maintaining this lib. :+1:

ziadoz:
Thanks, I've merged this.

PR #699, opened 2016-09-28
Add Studio

ivan1986:
A workbench for developing Composer packages.


josegonzalez:
How is this better/different from existing packages we list?

ivan1986:
In list not exist packages with same functions
it's replace some repositories in dev enviroment - very usefull if write own composer package.

josegonzalez:
This seems like the tooly package right under it, and some of it's functionality seems to be in composer already.

ivan1986:
Ok, i just suggest

josegonzalez:
Right. I'm asking you to justify your suggestion is all. Can you list a few ways this package is awesome?

PR #701, opened 2016-10-02
Add UnifiedArchive

wapmorgan:
I think it's important addition to your list.


josegonzalez:
Conflicts with master, needs a rebase.

Why do you think its an important addition?

wapmorgan:
proposed another pull #727

PR #703, opened 2016-10-04
Add metaphore (cache slam defense library).

sobstel:
Unique cache library to prevent a dogpile effect. 

It's complimentary solution to standard regular caching library (not replacement) for frequently requested and long running queries/code/etc.

Widely used in the wild for high-traffic websites like soccerway.com or goal.com.

More details: http://highscalability.com/blog/2014/7/30/preventing-the-dogpile-effect-problem-and-solution.html


josegonzalez:
Is there a better section to place this in?

sobstel:
Why do you think `caching` is not good section?

It doesn't fit any other category to me. It's caching-related, and solves one particular problem when cache gets expired, it takes relatively long to refresh data and many requests make your database/server hammered, which - like a snowball - makes it eventually down.

PR #704, opened 2016-10-04
Added Phalcon to micro-frameworks

afbora:
Phalcon also be used as micro framework.

Phalcon Doc: https://docs.phalconphp.com/en/latest/reference/micro.html


josegonzalez:
So can CakePHP or Laravel, using *just* their routing layer code.

PR #708, opened 2016-10-15
Added graphql-php

vladar:
Added graphql-php library


vladar:
For those interested: Github recently released their [GraphQL API](https://developer.github.com/early-access/graphql/). It uses ruby implementation of GraphQL. This PR suggests PHP implementation. Other big names to come.

PR #711, opened 2016-10-21
Hoa\String has moved to Hoa\Ustring

Hywan:


jadb:
@josegonzalez, checking logs shows there was no description before the bot commented. How do you think it could be better handled? Reducing the required length or dropping the rule all together?

josegonzalez:
Hmm didn't see that there weren't any logs. Maybe if the bot could point to the original contents somewhere, that would be cool.

Hywan:
I didn't write any description before the bot sent a comment. The title was self-describing enought ;-).

jadb:
I agree. The bot should probably require description only for PRs that are adding new links, typos and updates should get a pass. 

I'll work on improving that.

Thanks @Hywan!

PR #717, opened 2016-11-09
add Docker to Infrastructure section

jaschweder:
Added Docker has infrastructure option

### DIFF
--- a/README.md
+++ b/README.md
@@ -848,6 +848,7 @@ Libraries to help manage database schemas and migrations.
 
 * [appserver.io](http://appserver.io/) - A multithreaded application server for PHP, written in PHP.
 * [php-pm](https://github.com/php-pm/php-pm) - A process manager, supercharger and load balancer for PHP applications.
+* [Docker](https://www.docker.com/) - A containerization platform.
 
 # Resources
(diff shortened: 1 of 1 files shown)


jadb:
Sorry, not sure how that fits here. There is nothing PHP specific (compared to the other in same category).

jaschweder:
@jadb there are other tools in another sections, like DBV in Web Applications, Ansible in Development Enviorment, all tools in Deployment section, etc, that didn't are specific for PH, so if this is a rule here, please remove all those.

Otherwise, Docker is a generic tool that can be use to develop PHP applications, for a lot of uses cases.

josegonzalez:
@jaschweder Docker isn't infrastructure. If anything, it should be in Development Environment.

jadb:
@jaschweder sorry I maybe could have been clearer. I meant not a good fit for that specific category, not for the list as a whole.

jaschweder:
@jadb thanks for clarify :) I'll change the category of Docker to "Development Enviroment" as suggested

jaschweder:
done

jadb:
thanks @jaschweder!

PR #719, opened 2016-11-11
Added Swagger api documenter

funkyfundev:
Hi Guys,

I've added swagger for api documentation generator and JSON schema validator for Misc.


dereuromark:
I would prefer we merge the main pr first before doing others

alexkart:
> I would prefer we merge the main pr first before doing others

I agree; I think you can merge it already.

PR #721, opened 2016-11-14
Add IceHawk to list of micro frameworks

hollodotme:
IceHawk is a micro framework designed to build web applications and APIs following the CQRS (Command Query Responsibility Segregation) approach. It is stricly typed and written for PHP >= 7.0. https://github.com/icehawk


hollodotme:
The DNS issue with plugins.cakephp.org was solved (https://twitter.com/savant/status/798587996245295104), the build should pass now.

PR #724, opened 2016-11-21
Add Disco to Dependency Injection section.

shochdoerfer:
Disco is a container-interop compatible, annotation-based dependency injection container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,7 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 * [Acclimate](https://github.com/jeremeamia/acclimate-container) - A common interface to dependency injection containers and service locators.
 * [Auryn](https://github.com/rdlowrey/Auryn) - A recursive dependency injector.
 * [Container](https://github.com/thephpleague/container) - Another flexible dependency injection container.
+* [Disco](https://github.com/bitExpert/disco) - A container-interop compatible, annotation-based dependency injection container.
 * [PHP-DI](http://php-di.org/) - A dependency injection container that supports autowiring.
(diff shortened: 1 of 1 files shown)


josegonzalez:
How is this better/different than existing solutions on our list?

shochdoerfer:
Disco uses PHP code - PHP classes to be precise - as a configuration format. The methods of the class define the dependencies available. This is pretty unique approach and offers the following advantages over any other configuration format (e.g. XML, YAML, JSON, ...):
- Autocompletion and refactoring support with any PHP IDE
- Strict type checks within the configuration (e.g. for method return values as well as dependencies)
- Syntactically valid PHP class
- Really little code to write, just a method body and one annotation

So far I haven't seen any other DI implementation that makes use of classes as configuration format. A few other containers use plain PHP code as a configuration format. All those other solution suffer the same problem: At some point somewhere you do not depend on a concrete instance but on a string which is used to identify one of your dependencies. This means you cannot have a true type check in your configuration.

In the Disco configuration format this is different: When your method return value is wrong your IDE will show you an error. When the type of one of your dependencies does not match the expectation your IDE will show an error. No need to run into issues during runtime, errors can be spotted upfront when writing the configuration code. This is what makes Disco unique in its design.

mablae:
@shochdoerfer Sounds "awesome" !

shochdoerfer:
@josegonzalez anything else needed to drive things forward?

kalessil:
@shochdoerfer: looks awesome, really love "no-magic" approach here.

SenseException:
@shochdoerfer A good usage of PHP 7, which also means that improvements in PHP will also be part of Disco, right?

ziadoz:
Thanks, I've merged this as it seems sufficiently different to the existing solutions we've got on the list.

PR #732, opened 2016-12-03
Use https:// prefix for redis.io to fix failing build of Url Check

mablae:
…(redirect is marked as failure)


ziadoz:
Thanks for the fix.

PR #733, opened 2016-12-06
Add Wandu HTTP to HTTP section

wan2land:
wandu http provides session and cookie within psr7.


wan2land:
mistake! wrong url , sorry,

PR #735, opened 2016-12-13
Symfony version updated

greg606:
Changed from 2 to current 3


ziadoz:
Thanks for the update.

PR #736, opened 2016-12-20
Added Porter to Third Party APIs section.

Bilge:
Added [Porter](https://github.com/ScriptFUSION/Porter) to Third Party APIs section.

Unlike other entries in this category, Porter does not attempt to access a specific API, but rather, provides a framework for structuring code and importing data from any API.


geekish:
There is a separate [API section](https://github.com/ziadoz/awesome-php#api) for "libraries and web tools for developing APIs" that this might fit better in, but I feel like it would be a better fit under [Data Structure and Storage](https://github.com/ziadoz/awesome-php#data-structure-and-storage).

josegonzalez:
Agree with @geekish.

Bilge:
I don't feel strongly about how it should be categorized. If there is consensus on where to move it to I will change it accordingly.

josegonzalez:
Given that I have commit bit, I'll say you should go ahead and move it ;)

Bilge:
As requested, I have moved Porter to the Data Structure and Storage section.

PR #742, opened 2016-12-21
Remove deprecated assetic from the recommendations

MatthiasKuehneEllerhold:
Remove assetic from the recommendations because its maintainers marked it as deprecated.
See https://github.com/kriswallsmith/assetic/issues/79#issuecomment-266771392 for the details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -443,7 +443,6 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 ## Asset Management
 *Tools for managing, compressing and minifying website assets.*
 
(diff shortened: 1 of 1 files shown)


ziadoz:
@MatthiasKuehneEllerhold I read through the comments and didn't see anything that explicitly stated Assetic is deprecated, but it does sound like things are in maintenance mode, so I'll merge this in. I also did a quick run through of the libraries in the same category and removed Pipe, as it hadn't been updated since 2014. Thanks for submitting this.

garek007:
Assetic is honestly more trouble than it's worth. Damn bundle kept breaking on me and I found myself spending all of my time tracking down what was wrong with it. In the end I just used gulp to minify and combine my styles and scripts. Much less hassle.

PR #743, opened 2016-12-21
Add yasumi

MatthiasKuehneEllerhold:
Add yasumi for calculating national holidays.


ziadoz:
Thanks, this looks really useful.

PR #747, opened 2016-12-24
add php extension swoole

toxmc:
Swoole is an event-driven asynchronous & concurrent networking communication framework with high performance written only in C for PHP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,7 +469,6 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 
 ## Event
 *Libraries that are event-driven or implement non-blocking event loops.*
-
 * [Amp](https://github.com/amphp/amp) - An event driven non-blocking I/O library.
 * [Broadway](https://github.com/broadway/broadway) - An event source and CQRS library.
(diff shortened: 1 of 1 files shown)


smalleyes:
Why Swoole?

Traditional PHP applications almost always run behind Apache/Nginx, without much control of the request. This brings several limitations:

All memory will be freed after request. All PHP code needs be re-compiled on every request. Even with opcache enabled, all opcode still needs to be re-executed.
It is almost impossible to implement long connections and connections pooling techniques.
Implementing asynchronous tasks requires 3rd party queue servers, such as rabbitmq and beanstalkd.
Implementing realtime applications such as chatting server requires 3rd party languages, nodejs for example.
This why Swoole appeared. Swoole extends the use cases of PHP, and brings all these possibilities to the PHP world. By using Swoole, you can build enhanced web applications with more control, real-time chatting servers, etc more easily.

Gemorroj:
So, Swoole looks good. But he has a bad/outdated documentation.

smalleyes:
Swoole documents will get better

smalleyes:
lol.  this suggest can be merged？ why not.

PR #751, opened 2017-01-07
Add secure headers

BePsvPT:
[Secure Headers](https://github.com/BePsvPT/secure-headers) is a package that allow you to add security related headers to HTTP response.

The package is inspired by [twitter/secureheaders](https://github.com/twitter/secureheaders).


BePsvPT:
@josegonzalez could you please review this pull request?

josegonzalez:
This should be in the security section if anything. Please resubmit against that.

PR #758, opened 2017-01-31
Updates to Aura & Radar; adds PDS.

pmjones:
Changes are:

- Added topic-specific Aura libraries.
- Linked Aura Framework directly.
- Added Aura as a component library.
- Added PHP-PDS website.
- Moved Radar to micro-frameworks.
- "Typo" fix on MLAPHP ("moderni(s|z)ation")

Please let me know if these are appropriate or not, and thanks for putting together this list in the first place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 ## Frameworks
 *Web development frameworks.*
 
-* [Aura PHP](http://auraphp.com/) - A framework of independent components.
+* [Aura Framework](http://auraphp.com/framework/) - A framework built from independent components.
 * [CakePHP](https://cakephp.org/) - A rapid application development framework (CP).
 * [Laravel 5](https://laravel.com/) - Another PHP framework (L5).
 * [Nette](https://nette.org) - Another framework comprised of individual components.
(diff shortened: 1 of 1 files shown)


josegonzalez:
Do we really need to call out aura under each of the sections? I think one at the framework level is good enough.

pmjones:
@josegonzalez Maybe, maybe not. Aura's a funny bird, in that it is first-and-foremost a library collection; the framework is composed secondarily of those libraries. (As precedent, you might find that Cake shows up a few times.)

josegonzalez:
Interesting. I'd say that unless your solution is drastically better/different than existing ones, you shouldn't re-add it. Intl is a good example of that (both aura and cake are included) where there aren't many solutions and the two there are different enough from each other that they both warrant inclusion.

pmjones:
> I'd say that unless your solution is drastically better/different than existing ones

Of course, I think they are, but then I would; I invite you to review them yourself and judge accordingly.

josegonzalez:
Ugh more work. I'll try and get to this tonight (and a few other prs...)

pmjones:
Sure; I'm in no hurry. (I'll also add that the Aura libraries in some cases predate the other libraries, which may or may not factor into whether they should be included or not; e.g. Aura.View vs Foil, Plates, etc.)

pmjones:
@josegonzalez et al -- any further thoughts on this?

josegonzalez:
Fuck it.

PR #776, opened 2017-03-13
Add SideCI

akito19:
SideCI is an automated code reviewing tool with GitHub. It enables developers to speed up their code review and development cycle to improve productivity without requiring their to spend valuable time.


alexkart:
No longer available.

PR #778, opened 2017-03-26
Add stacker

Maxlab:
I've added stacker to list

Stacker - The environment for local web development on Docker.

#### Why stacker?
Stacker - This is a local environment for web development with everything you need. What is its benefit?
1. You do not need to manually configure the web server and add to the hosts, just cloned it and immediately launched it in the browser. It looks so [demo](https://youtu.be/42BemUfK5-4)
2. Inside, there is already everything that is needed in 90% of all cases. And if not, we will add it for you)
3. For you, there is a super zsh console with autocomplete and everything you need. Video with [presentation](https://youtu.be/N7HpPoNcaA4?list=PLD8VGB8i9TYha8YOd-deV6bX5hZco0ZGy) 
4. There is an autocompletion for Symfony and Laravel commands out of the box. For example, la5 and a double tab, will output a list of commands for which you can "walk" with arrows to select them
5. It is faster analogs, the same homestead is just a turtle compared to it
6. There is a [video course](https://www.youtube.com/playlist?list=PLD8VGB8i9TYha8YOd-deV6bX5hZco0ZGy)
7. Friendly author, in case there are questions or suggestions
8. Based on Docker. Wherever you can install Docker, you can install and Stacker
9. It is very simple to expand. The process of adding your own images with a couple of lines in docker-compose.yml
10. Just try it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -804,8 +804,9 @@ Libraries to help manage database schemas and migrations.
 * [Protobox](http://getprotobox.com/) - Another web tool for building PHP development virtual machines.
 * [PuPHPet](https://puphpet.com/) - A web tool for building PHP development virtual machines.
 * [Puppet](https://puppet.com/) - A server automation framework and application.
+* [Stacker](https://maxlab.github.io/stacker/) - The environment for local web development on Docker.
 * [Vagrant](https://www.vagrantup.com/) - A portable development environment utility.
-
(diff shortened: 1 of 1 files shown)


josegonzalez:
Sorry, this needs to be rebased to include just your change.

PR #786, opened 2017-04-14
CodeIgniter framework added to the list

mrNo0b:
CodeIgniter must be listed here.


josegonzalez:
Why must it be listed here?

josegonzalez:
(I realize thats a pedantic remark, we just have to hold every entry to the same standards).

dereuromark:
Dup of https://github.com/ziadoz/awesome-php/pull/693 it seems

ziadoz:
Thanks for the suggestion.

PR #795, opened 2017-05-11
Remove True Random

kelunik:
I don't see why this should be awesome. PHP offers a CSPRNG and mt_rand and covers all use cases well.


golaod:
Because both CSPRNG and mt_rand are **pseudo** random solutions, whereas random.org is free API which delivers truly random results.

kelunik:
@golaod Computers can't generate truly random numbers, so what's the point? What's the use case? What's the advantage?

juliangut:
Not a fan of making external requests for (pseudo-)random numbers @kelunik @josegonzalez but to honor randomness... 

from random.org (bold text is mine):

> RANDOM.ORG offers true random numbers to anyone on the Internet. The **randomness comes from atmospheric noise**, which for many purposes is better than the pseudo-random number algorithms typically used in computer programs.

kelunik:
@juliangut @golaod Feel free to resubmit it to another category like https://github.com/ziadoz/awesome-php#third-party-apis, but it definitely doesn't fit into Security.

PR #798, opened 2017-05-15
Added TCPDF PDF generation library.

veganaize:
TCPDF library for generating PDFs via PHP:
[https://tcpdf.org/](https://tcpdf.org/)


veganaize:
I don't think that this commit has anything to do with the Travis CI build failing.  It looks like that problem is to do with expired certs.

alexkart (approved):

PR #804, opened 2017-05-27
Added the ByJG Cache Engine class

byjg:
- A multi-purpose cache engine PSR-6 and PSR-16 implementation with several drivers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -696,6 +696,7 @@ Libraries to help manage database schemas and migrations.
 
 * [Alternative PHP Cache (APC)](http://php.net/manual/en/book.apc.php) - Open opcode cache for PHP.
 * [APIx Cache](https://github.com/frqnck/apix-cache) - A thin PSR-6 cache wrapper to various caching backends emphasising cache tagging and indexing.
+* [ByJG Cache Engine](https://github.com/byjg/cache-engine-php) - A multi-purpose cache engine PSR-6 and PSR-16 implementation with several drivers.
 * [CacheTool](https://github.com/gordalina/cachetool) - A tool to clear APC/opcode caches from the command line.
 * [Cake Cache](https://github.com/cakephp/cache) - A caching library (CP).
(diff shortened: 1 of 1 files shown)


byjg:
This library implements a several PSR-16 cache engines:
- FileSystem
- Memcached
- Redis
- Shared Memory (shmop)
- Array
- Dummy (do nothing)

Also enable you to create a PSR-6 compatible implementation by receiving **any** PSR-16 instance.

prisis:
And what is the difference from https://github.com/php-cache/cache

byjg:
The major difference is that this library have **BOTH PSR-16 and PSR-6** implementation and PHP Cache implements only PSR-6. Besides this cache-engine library is smaller than PHP Cache and is ideal for people who loves micro-frameworks.  

PSR-16 is a new PHP FIG standard called Simple Cache, that is less verbose than PSR-6. 

Other great feature is that the **CachePool** call existing in cache-engine can transform any PSR-16 implementation in a PSR-6 just doing:

```php
<?php
$cachePool = new CachePool($somePSR16Instance); // $cachePoll instance is a full PSR-6 implementation
```

byjg:
Sorry, I did not see in the docs how to implement a PSR-16 cache using PHP-Cache. I found right now the SimpleCacheBridge. But PHP-Cache and Cache Engine do the same thing with different approach. PHP-Cache creates a bridge from PSR-6 to SimpleCache, and Cache Engine creates a Bridge from SimpleCache to PSR-6. As SimpleCache is simplier to implement and is not as "fat" as PSR-6 there is a little advantage in the Cache Engine. 

But ALL is PSR Standard, and because this there is no difference in the interface. Everythink is CachePool and SimpleCache. So, We have to focus on this different approach. The Symfony/Cache do the same thing that PHP Cache, in the other others have a PSR-6 implementation and creates a bridge to PSR-16. The Cache Engine was fully refactored to create a PSR-16 (much simplier) and creates a Bridge to PSR-6.

PR #807, opened 2017-05-27
Added the ByJG Jwt Session

byjg:
- JwtSession is a PHP session replacement. Instead of use FileSystem, just use JWT TOKEN. The implementation follow the SessionHandlerInterface.


byjg:
This library is useful to create application stateless with zero effort to setup a new memcached or redis session server. There is no change in the code at all and there is no point of failure.

PR #815, opened 2017-06-07
Add Laravel components (Illuminate)

jleeothon:


sjelfull:
My main problem with this is that the components themselves isn't well documented, and it's hard/near impossible to use them outside of Laravel.

jleeothon:
@sjelfull I see, I'd agree is a concern and indeed it one of the things that I'm not too happy about while I was choosing some library for i.e. what `illuminate/support` does, although in the end this was the one I chose.

ziadoz:
I agree with @sjelfull. We already have Eloquent in the list, which does work well standalone, and is reasonably documented, I think that's enough here.

PR #816, opened 2017-06-07
Added latitude sql query builder

kfreiman:


kfreiman:
close it, becouse latitude does not support php5.5

PR #817, opened 2017-06-11
Update KNPRadBundle to KNP Rad Components

Einenlum:
KNPRadBundle is now outdated. It was split into several atomic components.


ziadoz:
Thanks!

PR #822, opened 2017-06-16
Adds hollodotme/fast-cgi-client to list of event libraries

hollodotme:
Fast CGI Client calls PHP scripts directly through the php-fpm socket (network or unix domain) and therefore implements the [FastCGI protocol](http://www.mit.edu/~yandros/doc/specs/fcgi-spec.html). Requests can be sent synchronously or asynchronously. In case of asynchronous requests it is possible to get the responses in various ways, blocking and non-blocking, ordered by request-order or reactive ordered by response time of the requests. It allows non-blocking loop integration and callbacks for responses and/or failures.


rabbl:
+1

alexkart (approved):

PR #824, opened 2017-06-23
A nicely useful scalable and easily deployable framework

jmurowaniecki:
hook

**HOOK is a restful, extendable backend as a service** that provides instant backend to develop sites and apps faster, with dead-simple integration with ios, android, javascript and more.

Was developed by a brilliant team and I am pretty sure that it will be very useful for the community.


josegonzalez:
Hasn't had any active development in over a year, so unfortunately I'll have to close this.

PR #828, opened 2017-07-06
Add new lib to debug session

tacnoman:
Adding lib to debug PHP applications in terminal, without any IDE. Is equivalent to ipdb in python or byebug in ruby.


tacnoman:
Adding lib to debug PHP applications in terminal, without any IDE. Is equivalent to ipdb in python or byebug in ruby.

tacnoman:
I opened again with a description. Is better make another merge request?

PR #837, opened 2017-08-10
Adding Gson

natebrunette:
This is a library for serializing and deserializing PHP objects. It takes advantage of PHP 7 language features to limit the number of annotations required for type information.

Additionally, it adds support for making (de)serialization decisions based on runtime information and provides a good interface for handling complex custom serializations.

https://github.com/tebru/gson-php


alexkart:
Doesn't seem to be maintained.

PR #838, opened 2017-08-11
added not mantained warning in Lock lib

odahcam:
as the bot asked for: added not mantained warning in Lock lib


odahcam:
explaned!

josegonzalez:
Or just delete it.

ziadoz:
I've deleted it from the list.

PR #839, opened 2017-08-17
Add Laradock

Mahmoudz:
[Laradock](https://github.com/laradock/laradock) is a full docker PHP development environment.
Used by nearly 3000 PHP developer as of today on a daily basis.


ziadoz:
Thanks, I merged a more recent PR with this as I hadn't realised it was a duplicate.

PR #840, opened 2017-08-22
Added projects of mine

laruence:
I think it's good to let more people know these tools.

thanks


ziadoz:
Thanks for the suggestions. We don't have a C-extension section, and there are no plans to add one for the foreseeable future, so as it stands I don't think these are a good fit. Sorry.

PR #845, opened 2017-09-06
Add Editorconfig-Checker

mstruebing:
https://github.com/editorconfig-checker/editorconfig-checker.php


mstruebing:
I think if the build is rerun, travis will run through.

PR #849, opened 2017-09-22
Add SonarQube platform

neyronius:
A top-quality tool that we use in our company to prevent potential errors.


alexkart:
The source repo has been deleted. Create a new PR if you think it is worth adding to the list.

PR #851, opened 2017-10-02
add QueryList scraper

jae-jae:
QueryList is a simple, elegant, extensible PHP Web Scraper .


dereuromark:
Does not look maintained anymore.

PR #860, opened 2017-10-19
Add Brick\Math and Brick\Money

BenMorel:
- [Brick\Math](https://github.com/brick/math) offers an implementation of large integer, decimal and rational numbers. It works with GMP, BCMath, and can even work without these extensions thanks to a pure PHP calculator.
- [Brick\Money](https://github.com/brick/money) is based on `Math` and provides exact calculations on monies of any size, with support for roundings, cash roundings, money contexts, currency conversion, and more.


BenMorel:
Closing in favour of #861 and #862 as suggested by VarCI.

PR #869, opened 2017-10-31
Moved Porter to architecture section and updated description

Bilge:
I'd like to propose reorganization of Porter under the architecture section as I feel it is a much better fit there. Its main purpose is to provide structure through its abstraction for data imports. This isn't the same thing as *data structure*, it's an *architectural* structure because we're more concerned with structuring the program and less about the structure of data. This is especially true of the upcoming [v4](https://github.com/ScriptFUSION/Porter/pull/46) which removes or migrates almost all implementations and replaces them with interfaces, but I have a hard time justifying its current classification even now.


ziadoz:
Thanks.

Issue #842, opened 2017-08-24
Library suggestion: Fuse


Loilo:
I'm opening this issue because I want to suggest a library (in which I'm certainly guilty of self-promision) but not start a pull request right away because I'm not sure how requirements should be interpreted in this case.

This is about my library [Fuse](https://github.com/Loilo/Fuse). It is a 100% API-compatible, almost line-by-line PHP port of [Fuse.js](http://fusejs.io/), which is a very popular fuzzy search library for Node.js and the web — those of you who are around in Frontend development may know it.

Some time ago, I was in need of a fuzzy search library, but there does not seem to be anything implemented in PHP that goes beyond some Levenshtein Distance fiddling. The results of those libraries weren't convincing to me, that's why I ported Fuse.js.

...therefore, regarding the requirements, the library surely "[is] unique in its approach or function" and "[fills] a niche gap in the market", however it is itself not well-known.

Thus "[being] widely recommended regardless of personal opinion" and "well known or discussed within the PHP community" does not fit the ported library, but it surely fits the JavaScript original.

What do you think about this?

geekcom:
Hey @Loilo great lib, congratulations.

Loilo:
Thanks – however none of the credit goes to me, it's all just Fuse.js. :)

ziadoz:
@Loilo It'd be good if you could open a proper PR for this, thanks.

PR #872, opened 2017-11-24
Add Fuse search library

loilo:
As requested in #842 🙂


alexkart:
The source repo has been deleted. Create a new PR if you think it is worth adding to the list.

PR #874, opened 2017-11-24
Library suggestion: MakeAccessible

e200:
Lightweight PHP package that let you easily access and test singleton classes and inaccessible instance members like private properties and protected methods.


e200:
@ziadoz

e200:
The command "awesome_bot README.md --white-list igor.io,symfony,toranproxy.com,vagrantup.com,3v4l.org,voicesoftheelephpant.com" exited with 1.

PR #882, opened 2017-12-06
Add HTTPlug to HTTP libraries

jdecool:
Recreate PR #825 because I've delete the origin fork.

This aim to add HTTPlug (an HTTP client abstraction without binding to a specific implementation) to HTTP libraries


ziadoz:
Thanks.

PR #889, opened 2018-01-19
Adding new dev env: devilbox

cytopia:
This PR will add another PHP Development Environment: [devilbox](https://github.com/cytopia/devilbox).

> The devilbox is a modern and highly customisable dockerized PHP stack supporting full LAMP and MEAN and running on all major platforms. The main goal is to easily switch and combine any version required for local development. It supports an unlimited number of projects for which vhosts and DNS records are created automatically. Email catch-all and popular development tools will be at your service as well. Configuration is not necessary, as everything is pre-setup with mass virtual hosting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -838,6 +838,7 @@ Libraries to help manage database schemas and migrations.
 * [Puppet](https://puppet.com/) - A server automation framework and application.
 * [Vagrant](https://www.vagrantup.com/) - A portable development environment utility.
 * [Docker](https://www.docker.com/) - A containerization platform.
+* [Devilbox](https://github.com/cytopia/devilbox) - A dockerized and general-purpose LAMP/MEAN stack for every PHP version.
 
 ## Virtual Machines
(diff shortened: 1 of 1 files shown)


josegonzalez:
How is this better or worse than any of the existing solutions?

cytopia:
This is not better or worse than the existing solutions. It is a completely different approach. The devilbox already gives you a fully functional development environment without you having to set up anything. The main benefit is that you can always switch between PHP 54, 55, 56, 70, 71, 72 and HHVM.

It will also allow you to have unlimitted projects without having to setup vhost configurations

cytopia:
@josegonzalez is this going to be merged or closed?

dereuromark:
If this gets rebased I think it can be merged 
I use it too and it is quite nice to be used

dereuromark:
Never mind, doesnt seem maintained anymore, stopped at 8.2.1

PR #893, opened 2018-02-02
Fix project name of ReactPHP

jsor:
Well, there is this other React, so we're always using [ReactPHP](https://github.com/reactphp/branding/#naming).

P.S. In case anyone is wondering, we had the name first (https://github.com/reactphp/react/commit/e36fb2623b0eb1018ce622bc0ba374a85be1788a vs. https://github.com/facebook/react/commit/75897c2dcd1dd3a6ca46284dd37e13d22b4b16b4) 😉


ziadoz:
Done, thanks.

PR #923, opened 2018-05-20
Add php examples from Martin's "Clean Code" book

ericpoe:
A member of my PHP UG showed this today. It's not complete, but it does show quite a few examples from Robert Martin's "Clean Code" book in PHP, even though the book's examples are in Java.


ericpoe:
The build fails are with links outside of my update.

ericpoe:
@ziadoz This should be ready for you now. Cheers!

PR #927, opened 2018-06-12
Add jwt library as a c extension

cdoco:
Add JSON Web Toke (JWT) library as a c extension.


ziadoz:
Thanks for your suggestion.

PR #931, opened 2018-07-03
Update with Brew PHP Switcher. Remove deprecated PHP tap.

peterlozano:
Brew PHP Switcher is the most currently maintained tool for switching between versions of php.


rybakit:
https://github.com/ziadoz/awesome-php/pull/652 can be closed.

PR #932, opened 2018-07-19
Added PeachPie to list of VMs

bfistein:
Added [PeachPie compiler](https://github.com/peachpiecompiler/peachpie) to the list of alternative VMs. I felt that it was the most appropriate place to put it, although it is not really a VM, but rather a complete runtime replacement.


ziadoz:
I'd be happy to add this if you can tweak the description to just say what it does more succinctly.

bfistein:
Shortened the description in the latest commit.

PR #934, opened 2018-07-27
Add PHPThemis, cryptographic library

vixentael:
[Themis](https://github.com/cossacklabs/themis) framework is cryptography library for making typical encryption schemes easy to use in the real-world use cases like data at rest, authenticated data exchange, transport protection, authentication, and so on.

[PHPThemis](https://github.com/cossacklabs/themis/wiki/PHP-Howto) is part of Themis suite for PHP applications, it has its own documentation, tutorials and examples.

Themis is made by cryptographers for applied developers: providing strong security, yet hiding complicated crypto-choices from developers. It has a lot of "pre-baked" use-cases: special encryption schemes for storing data and for sending data (f.e. no need to rotate transport keys manually; built in forward secrecy).


dereuromark:
Based on https://docs.cossacklabs.com/themis/languages/php/ and PHP 7.2 as highest supported it seems EOL for us here.

PR #939, opened 2018-08-27
Added codiad to IDE

mitchellurgero:
Codiad is a PHP/Web based IDE built for developing websites, scripts, and other web based software.

It is compatible with many languages including PHP, JS, HTML, XML, SH, BASH, BAT, and more.


mitchellurgero:
Updated.

PR #963, opened 2018-10-27
Remove Wercker

donis:
Wercker was acquired by Oracle and is being defunct in favour of their own product - Container Pipelines.


ziadoz:
Cheers.

PR #964, opened 2018-10-27
Add PHP section of awesome-podcasts

donis:
[Awesome podcasts](https://github.com/rShetty/awesome-podcasts#php) is a list of podcasts related to programming. They have a bit more and detailed information about PHP podcasts. 
It felt more right linking to the repo instead of copying info from it.


ziadoz:
@donis Thanks for the suggestion, but I'd prefer to just add PHP Podcasts directly to this list, as we already have a section for it. Nice list though. 👍

PR #966, opened 2018-11-10
Update README.md

pk-fr:
add yakpro-po:  a php obfuscator
framework agnostic, MIT license.


pk-fr:
I do not know why "needs: explanation", as i already added a description...

ziadoz:
Thanks for your suggestion.

PR #970, opened 2018-12-28
Added Rubix ML

andrewdalpino:
Rubix ML is a high-level machine learning library that lets you build programs that learn from data using the PHP language


DrDub:
RubixML is now at version 2.0.1. With close to 2,000 commits by two dozen contributors it might be time to revisit its inclusion in the list.

dereuromark:
Conflict requires resolving.

dereuromark:
Should be reopened without conflict.

PR #975, opened 2019-01-18
Remove Collections

vladimino:
In my opinion, it makes sense to remove [Collections](https://github.com/italolelis/collections) from the list as this package is archived by the owner and is not maintained anymore.


vladimino:
@roukmoute let's not mix up these 2 thing, adding a new library should be a separate PR.

ziadoz:
Thanks, I've merged this @vladimino. @roukmoute, feel free to open a PR for Doctrine Collections, it looks like a suitable replacement.

PR #976, opened 2019-01-23
Fix broken links

ericpoe:
The latest PRs have been blocked due to broken links outside of the scope of those PRs. This attempts to fix existing HTTP 301 and HTTP 302 errors that are registered as errors in the Jenkins build.

This also moves HTTP sites that can be linked as HTTPS to HTTPS as an attempt to future-proof as existing HTTP sites are pushed to become wholly HTTPS.


ziadoz:
Thanks for fixing these.

PR #988, opened 2019-05-10
Fix http301 errors

ericpoe:
There were 4 HTTP errors returned by the latest travis-ci test. 1 (phpmd) is because the project is abandoned and currently looking for a maintainer; 1 (securingphp) is because the HTTPS cert expired, but is now fixed; the 2 others are true HTTP 301 errors.


ziadoz:
@ericpoe Thanks for sorting these out.

PR #991, opened 2019-06-06
Remove Propel ORM

funtaps:
https://github.com/propelorm/Propel2#status-of-the-project

> Our advice:
>
> Use Doctrine if you are looking for a long-term supported, full-featured (and difficult to learn) and stable ORM.
> Use easy to learn Eloquent for prototyping and smaller projects.
> Help us build Propel3, which aims someday to be both of the above.


ziadoz:
Good spot. Thanks, I've merged this.

PR #993, opened 2019-06-06
Replace LazyRecord with Maghead

funtaps:
https://github.com/corneltek/LazyRecord
> The fast ORM for PHP. (This is a history repo, please visit maghead/maghead)


ziadoz:
Is Maghead actively developed? It seems like there hasn't been a release or commit in two years. It might be better to just remove LazyRecord.

ziadoz:
I've removed LazyRecord.

PR #996, opened 2019-06-30
Fix CI failures

ackintosh:
`drupal.org` seems to be denying bot access. 👀 

The server returns 403 against "User-Agent: awesome_bot":

```sh
$ curl -D - -s -o /dev/null https://www.drupal.org -H "User-Agent: awesome_bot" | grep HTTP
HTTP/2 403
```

([`awesome_bot` is a user-agent text used in awesome_bot.](https://github.com/dkhamsing/awesome_bot/blob/a1911a8ee4575a6bd603117d983a2a12be8df209/lib/awesome_bot/net.rb#L13))

We can receive 200 if change the User-Agent to other client one (e.g. curl ):

```sh
$ curl -D - -s -o /dev/null https://www.drupal.org -H "User-Agent: curl/7.54.0" | grep HTTP
HTTP/2 200
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,6 @@ rvm:
 before_script:
   - gem install awesome_bot
 script:
-  - awesome_bot README.md --white-list igor.io,symfony,toranproxy.com,vagrantup.com,3v4l.org,voicesoftheelephpant.com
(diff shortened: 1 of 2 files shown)


ackintosh:
@ziadoz I've fixed the CI failures. Please review this PR. ✨

ziadoz:
Thanks for this @ackintosh.

PR #1019, opened 2019-10-14
add spatie/menu to navigation section

dsazup:
Used this library in a project recently and found it very nice to work with. Definitely recommended 👍 https://github.com/spatie/menu


ziadoz:
Thanks!

PR #1022, opened 2019-10-18
Add Deptrac to Static Analysis

pdelre:
https://github.com/sensiolabs-de/deptrac

> Deptrac is a static code analysis tool that helps to enforce rules for dependencies between software layers in your PHP projects.
>
> For example, you can define a rule like "controllers may not depend on models". To ensure this, deptrac analyzes your code to find any usages of models in your controllers and will show you where this rule was violated.


ziadoz:
Thanks.

PR #1033, opened 2019-10-31
Add Setup PHP GitHub Action

shivammathur:
Add Setup PHP Github Action
>Setup PHP with required extensions, php.ini configuration and composer in GitHub Actions. This action can be added as a step in your action workflow and it will setup the PHP environment you need to test your application.


ziadoz:
Thanks, this looks useful.

PR #1044, opened 2020-01-05
Add PHP Architecture Tester

carlosas:
PHP Architecture Tester is a static analysis tool to verify architectural requirements.
It provides a natural language abstraction to define your own architectural rules and test them against your software.


jadb:
Thanks 👍

PR #1045, opened 2020-01-10
Add JSON Machine

halaxa:
It has a different approach to parsing big JSONs. Simply iterates over root or any key (sub-key) of a JSON using php's built-in `foreach`.


jadb:
Thanks 👍

halaxa:
Mee too :)

PR #1052, opened 2020-01-28
add PHP Static Installation

phpstatic:
add PHP Static Installation


jadb:
Not sure what this does. I am closing for now as when I tried visiting the website, I only got the Index of the public directory.

PR #1083, opened 2020-03-04
Add PHP Code Playground

ConradSollitt:


jadb:
Sorry, not really a PHP website. More like a playground for testing a new PHP framework.

ConradSollitt:
Ok thanks for responding so fast. I wasn't sure if I should add it or not.

Yes, you are correct it is a playground for a new PHP framework but it's mostly designed in a very generic way to use, try, and learn PHP.

PR #1084, opened 2020-03-17
feat: Add ImageOrientationFix

JellyBellyDev:
Add a library for fix image orientation by exif data with thanks to method exif_read_data.


jadb:
Thanks but I don't think this is a fit for the type of libraries included here.

JellyBellyDev:
Ok! No problem!

PR #1089, opened 2020-04-10
Add Ecotone Framework

dgafka:
Add Ecotone Framework.

Overview:
The main idea behind Ecotone Framework is to separate business and integration logic. So the business logic is not influenced by technical problems.
Ecotone tries to achieve that by enabling message driven architecture in PHP and following architectural patterns based on DDD and CQRS.


jadb:
Great! Good luck with it but let's wait for it to mature a little more before being included in the list here. 

Thanks!

dgafka:
Fair enough, thank you for review :)

PR #1092, opened 2020-04-23
docs: add imgix-php to imagery libraries

frederickfogerty:
This PR adds imgix-php to the list of awesome components. This is an awesome library because it makes images rendered by it automatically more performant, and responsive! The repository has been around for 5 years, well-documented, commercially-supported, and installable with Composer.


jadb:
Thanks for submitting but from what I can tell, imgix is a paid service and the URL you are adding is to the documentation on the commercial website and not the library itself on Github for example.

And is the library beneficial to the general PHP public or just imgix clients?

frederickfogerty:
Hey! Yeah, you're totally right on all fronts. The library is just beneficial to imgix customers, but imgix has a very generous free tier which is useful for anyone that serves images ✨

PR #1094, opened 2020-04-29
Add Path framework

Wharley01:


jadb:
Thank you for your suggestion.

Let's wait for it to mature a little more and gather some more stars on Github.

Wharley01:
Thanks for your response, There is no way it would get adoption if people
don't know about it, this is the first step of putting it out, I don't
think it's always just about the stars

PR #1101, opened 2020-05-25
add bottomline

fey:


dereuromark:
https://github.com/maciejczyzewski/bottomline seems not currently maintained.
Also, I would assume most is covered by composer autoload these days.

Issue #1118, opened 2020-09-09
Add ‚Kirby CMS‘


S1SYPHOS:
[Kirby](https://github.com/getkirby) is a flat-file CMS with an active community built around the core team that's been around for a couple years now (since 2012).

> Kirby: the CMS that adapts to any project, loved by developers and editors alike

For more information, see the GitHub link above or its [official homepage](https://getkirby.com).

Thanks for maintaining this awesome list!

Cheers,
S1SYPHOS

S1SYPHOS:
PR at your service, good Sir!

PR #1120, opened 2020-09-10
Add Kirby CMS

S1SYPHOS:
See #1118 - would be great!


ziadoz:
Thanks for the suggestion!

PR #1121, opened 2020-09-11
Fix Spatie Components link url

iyaki:
The url of Components > Spatie Open Source was not correct anymore.
Changed ir to the new one working url.


ziadoz:
Thanks!

PR #1131, opened 2020-11-04
Add DecimalObject value object library

dereuromark:
has been battle tested the last year inside larger e-commerce applications.
PHP 8 ready


ziadoz:
Thanks!

PR #1132, opened 2020-11-14
fzaninotto/Faker is archived and won't be updated anymore.

skoro:
faker library is updated to the new repository.


ziadoz:
Thanks @skoro

PR #1137, opened 2020-12-31
Add 'PHP Number' to Numbers section

bobmulder:


dereuromark:
Is this still supported/active?
Or can this be closed in favor of existing others? https://github.com/ziadoz/awesome-php?tab=readme-ov-file#numbers

bobmulder:
Well in meantime it earned some stars. I still maintain the package so I'd be happy if it can be added to the list!

dereuromark:
It would be great if it would also run latest 8.3 in CI
That people know it works and is still great :)

bobmulder:
I actually use it in 8.3 but great idea to give it an update. Will do it in a few days. That you Mark!

PR #1140, opened 2021-02-10
Add info about Manticore Search

sanikolaev:
Sphinx as an open source project has been forked to Manticore Search and has been being actively developed since 2017


ziadoz:
Thanks!

PR #1146, opened 2021-03-26
Add Chamilo in CMS section.

stefleininger:
Add Chamilo in CMS section.

https://chamilo.org/


dereuromark:
Doesnt seem to be maintained

> Please note: For version 1.11.26, Chamilo only supports PHP 7.4.

PR #1149, opened 2021-05-15
Add Codeigniter to frameworks list

victory7:
Added Codeigniter framework.
https://codeigniter.com/
 
This framework is one of the most popular PHP frameworks.


kubawerlos:
OMG...

https://github.com/ziadoz/awesome-php/pulls?q=is%3Apr+Codeigniter+is%3Aclosed

victory7:
I guess these duplications are because no one guesses a PR added before and didn't merge or refused!
WHY?
CI is good, lightweight, and awesome for many small to large projects even in version 3. You can easily check the most popular PHP frameworks in 2021!
Your list has many useless and out-of-date libraries that maybe someones like and prefer to use. But you don't list a framework with thousands of users around the world.

PR #1172, opened 2021-11-26
Changed emoji library

EgorBakulin:
A library has more useful functionality and support latest versions of php


ziadoz:
Thanks for the suggestion.

PR #1175, opened 2022-01-04
Add leaf php to micro frameworks

lakuapik:
leaf


rkrx:
Why would someone use leaf who could also use Laravel?

The excessive use of static calls is really scaring me right now.

lakuapik:
@rkrx imo, its a user preference. maybe some people has a use for it, and some does not. it does not matter. what matter is this is one of php gems that i think awesome enough to be added on the list.

ziadoz:
Thanks for the suggestion.

justbyitself:
Why not leaf? It's an active projectc like Slim but lighter. Add it, please.

alexkart:
> Why not leaf? It's an active project like Slim but lighter. Add it, please.

Please create a new PR if you think it is worth adding, as this source repo has already been deleted.

PR #1178, opened 2022-01-23
Add `kubephp`

sherifabdlnaby:
Suggesting to add https://github.com/sherifabdlnaby/kubephp to Awesome PHP.

I understand that "self-promotion" is frowned upon, however, I found no similar project referenced here.

I put it under `Infrastructure` because it's as close to what `kubephp` is, which is being a Production base template for your docker image.

The second category that it might belong to is `Development Environment`, although `kubephp` helps with Development with the `dev` docker image, it's also a production docker image template and hence I didn't put it under 'Development Environment`.


ziadoz:
Thanks for your suggestion.

PR #1180, opened 2022-02-14
Fix `#files` link

chadidi:
Github taking to repo files instead of the reference down there.

### Reproduce by clicking this link
https://github.com/ziadoz/awesome-php/#files

image


ziadoz:
Good spot, thanks.

PR #1195, opened 2022-06-28
Update README.md

kandikits:
kandi kit link added under Resources. We publish the best of the libraries for PHP and hence recommend to include in your Awesome list. Thanks!


kandikits:
Its Unfortunate the Testing Libraries list didn't match the expectations.
How about https://kandi.openweaver.com/explore/php ?

PR #1210, opened 2022-12-12
Add `ddeboer/imap`

Slamdunk:
Object-oriented, fully tested PHP IMAP library

https://github.com/ddeboer/imap


dereuromark (approved):
Once the fullstop is added at the end.

Also, we should remove https://github.com/tedious/Fetch and https://github.com/barbushin/php-imap as it seems abandoned.
So this would be the only IMAP one left.

PR #1222, opened 2023-04-13
Added 100 PHP Mistakes and How to Avoid Them

stjepanjurekovic:
Hi,

Stjepan here from Manning. I thought this title might be a good match for your list. Thank you for considering it.

Best,


ziadoz:
Thanks for your suggestion.

PR #1241, opened 2023-12-28
Added 'Fat-Free Framework' in the Micro-Frameworks section.

AndiKod:
The project seems to be alive, actually v3.8 and a v4 in the making.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,7 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 ### Micro Frameworks
 *Micro frameworks and routers.*
 
+* [Fat-Free Framework](https://www.fatfreeframework.com/) - A powerful yet easy to use PHP micro-framework.
 * [Laravel-Zero](https://laravel-zero.com) - A micro-framework for console applications.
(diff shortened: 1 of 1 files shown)


oleg-andreyev:
> The philosophy behind the framework and its approach to software architecture is towards minimalism in structural components, **avoiding application complexity** and **striking a balance between code elegance**, application performance and **programmer productivity**.

IMO: No. No. No.

I’ve took over a legacy project with 3F and it’s impossible to implement something scalable and maintainable

AndiKod:
I understand that you had an experience that wasn't what you was expecting. It's more a matter of taste, habits and use-case. 

I proposed the link as in the field of PHP-MicroFrameworks FatFree are here since a while and used in the correct context it do does what it's supposed to do.

But once again, if the Awesome-PHP repo' is about listing the personal preferences of the maintainers, that my bad, won't bother contribute further.

Have an Awesome ...almost new year ;)

ziadoz:
Pretty sure this is a duplicate.

AndiKod:
> Pretty sure this is a duplicate.

Might be, indeed. All things considered, sorry for the noise, you're right.

PR #1261, opened 2024-03-01
Add Leaf micro framework

justbyitself:


ziadoz:
This has been suggested before. https://github.com/ziadoz/awesome-php/pulls?q=is%3Apr+leaf+is%3Aclosed

justbyitself:
but why not?

PR #1272, opened 2024-03-10
Remove Lumen micro framework

justbyitself:
Official documentation says: "we no longer recommend that you begin new projects with Lumen". https://lumen.laravel.com/docs/10.x#installation


SupianIDz:
no longer being recommended doesn't mean it's not needed by people.

Even though it is no longer officially recommended, Lumen is still supported today (at least until version 10 and security fix support until February 4th, 2025).

I don't agree with Lumen being removed.

PR #1294, opened 2024-07-05
Added LeafPHP to the micro frameworks section

maslukasz:
Added LeafPHP to the micro frameworks section

https://github.com/leafsphp/leaf


ziadoz:
This is a duplicate, and has been rejected before.

PR #1305, opened 2024-09-03
Update README.md

jeffmcneill:
Removed spress from static site generators, bad URL


alexkart:
Yes, it looks like it is no longer being maintained.

PR #1333, opened 2025-02-09
Added OpenMage to CMS

sreichel:
See https://www.openmage.org/

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 * [Kirby](https://getkirby.com/) - A flat-file CMS that adapts to any project.
 * [Magento](https://github.com/magento/magento2) - The most popular e-commerce platform.
 * [Moodle](https://moodle.org/) - An open-source learning platform.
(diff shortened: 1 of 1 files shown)


sreichel:
Bump, b/c it is awesome :=)

dereuromark:
I understand the need.
But I am not sure it qualifies as awesome under the guidelines 
E.g 
https://github.com/OpenMage/magento-lts/blob/82c06be915280f6989758fe241615c0c99cf1f02/composer.json#L10

Bc is important, sure. 
But then it might be outdated in terms of current version support.
I wonder why they didn't try to provide also 8.4+ support.

sreichel:
Current composer says ...https://github.com/OpenMage/magento-lts/blob/main/composer.json#L11

We are working on php8.x support all the time. That makes it awesome :P

dereuromark:
My bad. Must have opened the wrong branch.

sreichel:
OpenMage IS awesome, because its still alive after years becoming EOL.

We are also here since some weeks: https://ddev.readthedocs.io/en/stable/users/quickstart/#openmage
We are contributing to https://github.com/Shardj/zf1-future 

We are the older, but better Magento :)

alexkart (approved):

PR #1334, opened 2025-02-14
Web based POS

arshidkv12:


Mustafa23232 (approved):

dereuromark:
Why is this named Web POS when the repo and content says sth different, I wonder.

How are we with copyleft licenses by the way, compared to MIT etc.

PR #1335, opened 2025-02-20
Replace Symfony Security Monitoring

bebehr:
[Symfony Security Monitoring](https://security.symfony.com)  does not exist anymore.

[Local PHP Security Checker](https://github.com/fabpot/local-php-security-checker) is mentioned in Symfony docs as a tip for CI standalone-tool, see: https://symfony.com/doc/7.3/setup.html#checking-security-vulnerabilities

This PR replaces Symfony Security Monitoring with Local PHP Security Checker.


bebehr:
Fixes: https://github.com/ziadoz/awesome-php/issues/1336

alexkart:
It looks like Local PHP Security Checker is also archived and they recommend to just use `composer audit` instead.
I think we just need to remove Symfony Security Monitoring from the list.

bebehr:
Removed Local PHP Security Checker as replacement. So simply removes Symfony Security Monitoring. No need to add `composer audit` to list.

PR #1351, opened 2025-05-01
Add Cecil as static site generator

welcoMattic:
cc @ArnaudLigny


dereuromark:
should be in alphab. order, see guidelines

dereuromark (approved):

PR #1352, opened 2025-05-07
Add Hi.Events

daveearley:
Hello 👋 

First off, thank you for putting together and maintaining this list - I've discovered some great resources here over the years.

I'd like to propose adding [Hi.Events](https://github.com/HiEventsDev/Hi.Events), an open-source event management platform built with Laravel.

I couldn't find an existing category that suited Hi.Events, so I’ve added a new Event Management section. The closest match was E-commerce, but that felt too broad and not quite right for a dedicated event ticketing system.

Hi.Events is Composer-installable, actively maintained, targets recent PHP versions, and is well documented.

Let me know if any changes are needed.

Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,6 +555,11 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 * [Swap](https://github.com/florianv/swap) - An exchange rates library.
 * [Sylius](https://sylius.com/) - An open source e-commerce solution.
 
+### Event Management
+*Applications and tools for managing events, selling tickets, and handling attendees.*
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -555,6 +555,11 @@ Please see [CONTRIBUTING](https://github.com/ziadoz/awesome-php/blob/master/CONT
 * [Swap](https://github.com/florianv/swap) - An exchange rates library.
 * [Sylius](https://sylius.com/) - An open source e-commerce solution.
 
+### Event Management
```
dereuromark: Not sure if its own category makes sense, maybe this should be under https://github.com/ziadoz/awesome-php?tab=readme-ov-file#web-applications or alike? 
cc @alexkart

PR #1355, opened 2025-06-01
add garlic-hub - Digital Signage CMS compatible to SMIL

sagiadinos:


dereuromark:
Without releases this is most likely not being able to get added.
See also guidelines etc.

PR #1358, opened 2025-07-14
Update README.md

wemmbley:
Added new php framework.


dereuromark:
New frameworks are going to be harder to add, as they might not yet be "awesome".

PR #1363, opened 2025-08-23
Add pikaid to Strings section

larananas:
Add [Pikaid](https://github.com/pikaid/pikaid-php) to the Strings section.

Pikaid (pronounced “Pika ID”) is a 26-character, lowercase Base36 identifier designed to be:
  -Small: same length as ULID (26 chars) but more entropy-packed.
  -Sortable: lexicographically ordered by creation timestamp without extra fields.
  -Secure: 96 bits of cryptographic randomness to make collisions practically impossible.


larananas:
Hi! Just a gentle follow-up on this PR in case it fell through the cracks.
If there’s anything I should update or adjust to better match the project guidelines, I’d be happy to do it. Thanks for your time.

PR #1408, opened 2026-04-01
Add ShopSavvy

jakemarsh:
Added ShopSavvy's PHP SDK to the Third Party APIs section — it's a client library for the ShopSavvy product data and price comparison API.


rkrx:
Very limited audience

jakemarsh:
Fair point — thanks for the feedback and for maintaining such a great list.

PR #1463, opened 2026-07-17
Add Laravel Image Sanitize to filtering and sanitizing

mathiasonea:
## Summary
- add Laravel Image Sanitize to the Filtering, Sanitizing and Validation section
- include the package repository and a short description

## Project
https://github.com/laravel-at/laravel-image-sanitize

## Why
Laravel Image Sanitize is a focused PHP/Laravel package for detecting suspicious image upload payloads and re-encoding unsafe raster files. It is actively maintained, Composer-installable, documented in English, and addresses a specific upload-security niche rather than general validation.


dereuromark:
Why not adding this to a Laravel awesome list?
If u can extract an agnostic version it might fit better here.

mathiasonea:
That's actually a great idea. I'll work on an agnostic PHP version, and resubmit. Thanks!

